PR #6, opened 2016-09-29
Prevent user with empty uid

butonic:
The uid might be empty if the mapped owncloud_username, eg. mail is empty. While admins should make   sure the attribute is always set the ldap admins are only human. Errors happen and we need to protect against user errer. In this case we should die hard, giving everyone a hint at what might be wrong, which is why I use assert(). It will throw an exception and prevent further desaster. As in overwriting the quota for all users ...

### DIFF
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -4,6 +4,7 @@
  * @author Joas Schilling <coding@schilljs.com>
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
  *
  * @copyright Copyright (c) 2016, ownCloud GmbH.
  * @license AGPL-3.0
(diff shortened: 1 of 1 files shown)


butonic:
@owncloud/ldap quick review

jvillafanez:
🚫

Assertions can be disabled, and we should assume they will disabled on production. Taking into account that the code depends on an external service (LDAP) whose configuration is completely unknown, we can't consider that the code will be "safe" to be run without assertions once it reach production.

Just checking and throwing the exception seems a better option.

butonic:
@jvillafanez @DeepDiver1975  done, makes sense.

jvillafanez:
👍

PVince81:
triple combo facepalm hit combo

PVince81:
Please backport to 9.1 and 9.0

jvillafanez:
stable9.1: https://github.com/owncloud/core/pull/26810
stable9: https://github.com/owncloud/core/pull/26811

SergioBertolinSG:
This field corresponds in the ldap app's expert tab to uuid or internal username?

PR #7, opened 2016-09-29
Allow specifying the prefix for occ ldap:create-empty-config

butonic:
## Description

Adds an optional `configId` param to `occ ldap:create-empty-config`
## Related Issue

new feature
## Motivation and Context

It improves scriptability of ldap setups, eg for automated testing with multiple ldap servers
## How Has This Been Tested?

locally, juggling 7 ldap servers, writing a script that automatically sets up the connections
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
  The configuration class and user_ldap does not yet use DI ... makes it hard to write a unit test. oh well
- [ ] All new and existing tests passed.

### DIFF
--- a/lib/Command/CreateEmptyConfig.php
+++ b/lib/Command/CreateEmptyConfig.php
@@ -25,6 +25,7 @@
 namespace OCA\User_LDAP\Command;
 
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use \OCA\User_LDAP\Helper;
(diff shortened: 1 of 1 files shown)


PVince81:
@jvillafanez trivial review ? merge ?

lib/Command/CreateEmptyConfig.php
```diff
@@ -46,19 +47,33 @@ protected function configure() {
...
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
-		$configPrefix = $this->getNewConfigurationPrefix();
+		$availableConfigs = $this->helper->getServerConfigurationPrefixes();
+		$configID = $input->getArgument('configID');
+		if(is_null($configID)) {
+			$configPrefix = $this->getNewConfigurationPrefix($availableConfigs);
+		} else {
+			if(in_array($configID, $availableConfigs)) {
```
jvillafanez: I'd rather use always a strict comparison or add an explicit comment otherwise

lib/Command/CreateEmptyConfig.php
```diff
@@ -46,19 +47,33 @@ protected function configure() {
...
+				$output->writeln("configID already exists");
+				return;
```
jvillafanez: return `1` for the error condition?

jvillafanez:
Other than my comments, 👍

PVince81:
@butonic please check the minor issues. If no time, defer to "planned"

tomneedham:
Implemented JP's comments and added a check that we dont try to create an empty config whcih happened if you tried with `occ ldap:create-empty-config ''`

tomneedham:
Now you can either give nothing and you get given an id, or specify a string which is used if possible

lib/Command/CreateEmptyConfig.php
```diff
@@ -46,19 +47,39 @@ protected function configure() {
...
+		if(is_null($configID)) {
+			$configPrefix = $this->getNewConfigurationPrefix($availableConfigs);
+		} else {
+			// Check we are not trying to create an empty configid
+			if($configID === '') {
+				$output->writeln("configID cannot be empty");
+				return 1;
+			}
+			// Check if we are not already using this configid
+			if(in_array($configID, $availableConfigs)) {
```
jvillafanez: The `in_array` function has a parameter to specify strict type checking. I'd rather use that option always by default or add a comment otherwise explaining why it isn't needed. This should prevent unwanted surprises.

tomneedham:
Rebased with the CI fix so let's see it runs through

PR #8, opened 2016-09-30
Fix possible race condition when trying to connect to the server and …

jvillafanez:
…setting a new configuration value just before

The new value might not be saved in time when the connection is being
made

Original PR: https://github.com/owncloud/core/pull/25080 (fixing https://github.com/owncloud/core/issues/24303)


PVince81:
:+1:

PR #18, opened 2016-10-09
array_merge does not merge keys correctly

goodkiller:
When you are using user ID-s as numeric and using numeric ID-s in keys then array_merge does not merge correctly.
Reference: https://github.com/owncloud/core/issues/13378


PVince81:
@owncloud/ldap

jvillafanez:
@goodkiller Could you explain how did you manage to have numeric ids there?

DNs should be strings. Any member, memberuid or similar attribute should be strings. I see strings everywhere, and in that context it doesn't matter what code we use, the only difference is how duplicated DNs are handled, choosing either the first or the second, which in fact doesn't matter because we have another problem.

PVince81:
@jvillafanez could it be that PHP did its thing and converted "strings that look like int" to int ?

anyway, no feedback -> close

Issue #22, opened 2016-10-19
Wizard detecting memberUid doesn't save it


PVince81:
### Steps

1. Use the zombie army from https://github.com/owncloud/administration/tree/master/ldap-testing
1. Create users and groups with the PHP scripts
1. Edit "Box10" and add a few zombies as members using the "memberUid" attribute.
1. Setup LDAP in OC
1. On the main page, click "Next"
1. Then click "verify"
1. Then "Next" twice
1. In the groups page, pick "posixGroup", then click "Verify"
1. Then go to the Advanced page and open "Directory settings"
1. See that "memberUid" is auto-detected and set
1. Refresh the page
1. Go to the Advanced page and open "Directory settings"

### Expected result
"memberUid" setting sticks

### Actual result
The setting reset itself back to "uniqueMember".

### Versions
Seen on stable9.1 and master

@jvillafanez maybe some ajax race condition that doesn't save that setting properly after auto-detecting ?

PVince81:
In stable9 however, it directly shows "uniqueMember" and the value stays. So at least it matches whatever is on the server.

PVince81:
The same happens for other fields like email. In my setup it gets set to "mail" but when I refresh the page it disappears.

My workaround is when I arrive on this page with the autodetected values, I open the dropdown and close it again to force saving. Then I edit the email field, add a char then remove it. This saves it too.

So I guess there is some saving routine missing after auto-detecting.

PR #23, opened 2016-10-25
Add prefix to the connection

jvillafanez:
No linked issue that I'm aware.

@owncloud/qa we'll need to test this.

@PVince81 Not sure if we'll need to backport or not.


PVince81:
What is the prefix about ? What issues might appear without this fix ? How to test ?

jvillafanez:
1. Configure 2 LDAPs (use different configuration details)
2. Reload the page
3. Delete the first configuration
4. Reload the page (you might need to repeat this twice)

We'll have 2 configurations instead of one. The second is duplicated.

https://github.com/owncloud/user_ldap/issues/22 might also be affected by this.

PVince81:
:+1: for the fix.

However it doesn't solve the "memberUid" refresh issue

Issue #33, opened 2016-12-15
Autodetecting FreeIPA UUID field


xelez:
In file `lib/Access.php` (https://github.com/owncloud/user_ldap/blob/master/lib/Access.php#L1487)
```php
$testAttributes = array('entryuuid', 'nsuniqueid', 'objectguid', 'guid', 'ipauniqueid');
```

Here is everything ok, `ipauniqueid` is in the array of attributes to try.

But in file `lib/Connection.php` (https://github.com/owncloud/user_ldap/blob/master/lib/Connection.php#L340),

```php
$uuidAttributes = array('auto', 'entryuuid', 'nsuniqueid',
                                                           'objectguid', 'guid');`
```

`ipauniqueid` is missed from array, so `softValidation()` changes it back to 'auto' and so autodetecting breaks.

Adding `ipauniqueid` to the $uuidAttributes solves the issue.

Temporary solution: set UUID attribute in expert settings.

P.S. Maybe it should be not an issue but a pull request, but I'm not sure for something that small.

PR #34, opened 2016-12-16
autodetect 'ipauniqueid' in Connection as well

butonic:
fixes https://github.com/owncloud/user_ldap/issues/33

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1484,15 +1484,13 @@ class Access extends LDAPUtility implements IUserTools {
 			return true;
 		}
 
-		// for now, supported attributes are entryUUID, nsuniqueid, objectGUID, ipaUniqueID
-		$testAttributes = array('entryuuid', 'nsuniqueid', 'objectguid', 'guid', 'ipauniqueid');
-
-		foreach($testAttributes as $attribute) {
+		foreach($this->connection->uuidAttributes as $attribute) {
 			$value = $this->readAttribute($dn, $attribute);
 			if(is_array($value) && isset($value[0]) && !empty($value[0])) {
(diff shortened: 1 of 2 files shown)


butonic:
@xelez can you test?

xelez:
@butonic Tested, works. But I think at least a comment near both arrays, mentioning that you have change the other array is needed.

xelez:
Is it going to be merged?

PVince81:
> that you have change the other array is needed so that this error

@xelez not sure what you mean with your comment.

@jvillafanez is this PR acceptable ?

butonic:
IMO for now this makes sense ... the autodectection stuff may add additional queries, so you should configure it in the expert settings anyway.

xelez:
@PVince81 I mean, this bug emerged because you had to add 'ipauniqueid' into 2 different arrays.
And original author [have forgotten](https://github.com/owncloud/user_ldap/commit/c47b37e13d76073fc5283fbb6552bf29e72d0022) to add it to the second one.

I think it's a good idea to do something to prevent such errors in the future/ (add comment `// You should probably modify another array <here> as well!` ?)

P.S. Sorry for my English

PVince81:
Ah ok, understood. I wasn't aware that the array was duplicated. Maybe best would be to put it into a constant instead and use it in both locations.

jvillafanez:
👍

butonic:
@PVince81 @jvillafanez unified the code.

jvillafanez:
Nothing wrong with the tests. Either kick it or merge it.

jvillafanez:
Rebased

PR #17, opened 2016-10-05
Ldap integration tests 1up

jvillafanez:
Original PR: https://github.com/owncloud/core/pull/23832

**NOTE** https://github.com/owncloud/core/pull/23832/commits/acdb61ca6caf79a88567ecc25206ef1cda94eecc skipped (modifying the .gitignore file of the core repo).

PR #36, opened 2016-12-20
Added remaining test for user home

SergioBertolinSG:
There was a test removed from the original suite:
https://github.com/owncloud/user_ldap/pull/17

It is added here.

It fails because a search function doesn't return any users.


tomneedham:
@SergioBertolinSG do we still need this? How do we fix the failing test? Otherwise let's kill this :)

SergioBertolinSG:
I guess we can kill it, @individual-it is working on a new test suite.

tomneedham:
cc @phil-davis also

PR #37, opened 2017-01-11
Delete IntegrationFetchUsersByLoginName

davitol:
Delete IntegrationFetchUsersByLoginName.php file due to it is not needed. IntegrationTestFetchUsersByLoginName.php is fair enough.


davitol:
@SergioBertolinSG @PVince81

PR #38, opened 2017-01-13
set default quota when no ldap attribute is defined

butonic:
Without this a default quota is only eventually updated. It should be updated on login.

cc @dercorn

### DIFF
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -170,8 +170,11 @@ class User {
 
 		//Email
 		$attr = strtolower($this->connection->ldapEmailAttribute);
+		$quotaDefault = $this->connection->ldapQuotaDefault;
 		if(isset($ldapEntry[$attr])) {
 			$this->updateEmail($ldapEntry[$attr][0]);
+		} else if (!empty($quotaDefault)) {
+			$this->updateQuota();
 		}
 		unset($attr);
 


PVince81:
Please add unit tests.

PVince81:
Any update on the unit tests ?

PVince81:
@jvillafanez review ? (unit tests pending)

lib/User/User.php
```diff
@@ -170,8 +170,11 @@ public function processAttributes($ldapEntry) {
 
 		//Email
 		$attr = strtolower($this->connection->ldapEmailAttribute);
+		$quotaDefault = $this->connection->ldapQuotaDefault;
 		if(isset($ldapEntry[$attr])) {
 			$this->updateEmail($ldapEntry[$attr][0]);
+		} else if (!empty($quotaDefault)) {
```
jvillafanez: ⁉️ 

I expect to update the quota inside the quota section, not in the email section.

PVince81:
@jvillafanez is this obsoleted by your other PR that makes all quota behavior consistent ? Or is this an addition ?

jvillafanez:
This is obsolete from my point of view

PVince81:
the actual quota fix will be in 9.1.5 and 9.0.9

PVince81:
the actual quota fix will be in 9.1.5 and 9.0.9

Issue #52, opened 2017-02-15
Adjust settings page section


PVince81:
With the new settings page in 10.0, the LDAP settings appear under "Additional settings" which is not easy to find at all for admins.

The app needs to be adjusted to use the new interfaces (@tomneedham can you point to the matching developer documentation ?) in order to have a "LDAP" section in the left sidebar.

Setting this to high because I'm pretty sure admins will be very confused and frustrated.

PVince81:
bug as in "UX bug"

PVince81:
Too bad we don't have subsections, else we could in the future put the tabs of the wizard into several subsections

DeepDiver1975:
Question: do we need the wizard in 10?

PVince81:
We still need a way to configure the LDAP connection and various filters to know what and how to import stuff into the account table.

DeepDiver1975:
yes - most probably ... we will see

PVince81:
@tomneedham can you first prepare developer docs ? Then maybe other people would be able to help adjusting this. (for example @noveens)

PR #57, opened 2017-03-02
Use new settings registration method.

tomneedham:
Fixes https://github.com/owncloud/user_ldap/issues/52


PVince81:
I don't think "Authentication" is the correct term here. LDAP is not only used for authentication for also for sharing if you specify groups.
Not sure what would fit better:
- User backend
- User sources
- User providers
- ???

@jvillafanez any better ideas ?

PVince81:
Also maybe use a person icon instead of cog ?

PVince81:
Otherwise it looks good. The wizard works.

A small little detail: when enabling the LDAP app from the apps page, the "Authentication" section doesn't appear automatically. I need to refresh the page. Solving this is probably not that easy.

jvillafanez:
"User backends" sounds more technical. I'd choose "user providers".

tomneedham:
https://github.com/owncloud/core/pull/27296#issue-211383466 adds the icon for the 'authentication' panel. 'User Providers' it is then?

tomneedham:
> A small little detail: when enabling the LDAP app from the apps page, the "Authentication" section doesn't appear automatically. I need to refresh the page. Solving this is probably not that easy.

Correct. When we ajax-ify it all we can refresh the sections probably quite easily.

DeepDiver1975:
Generally speaking this works - anything else can be adjusted later :+1:

Issue #47, opened 2017-02-08
One step recursion for each LDAP attribute


PVince81:
Steps:

1. Setup LDAP users with quota, email and avatar fields (no memcache)
1. Login as a LDAP user B
1. Share a folder "test" with user A
1. Login as a LDAP user C
1. Share a folder "test2" with user A
1. Log out
1. `delete from oc_preferences where configkey='email' or configkey='lastFeatureRefresh' or configkey='quota';`
1. Put breakpoint at `OCA\User_LDAP\User\User->updateEmail() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:435
1. Login as user A
1. Step into the attribute updating code

### Expected result
No recursion at all

### Actual result
Recursion that calls `userManager->get()` for each attribute.
Fortunately the recursion is stopped by `wasRefreshed('email')`.

```
0  OCA\User_LDAP\User\User->updateAvatar() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:498
1  OCA\User_LDAP\User\User->update() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:143
2  OCA\User_LDAP\User_LDAP->userExists() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User_LDAP.php:277
3  OC\User\Manager->get() /srv/www/htdocs/owncloud/lib/private/User/Manager.php:139
4  OCA\User_LDAP\User\User->updateQuota() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:470
5  OCA\User_LDAP\User\User->update() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:139
6  OCA\User_LDAP\User_LDAP->userExists() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User_LDAP.php:277
7  OC\User\Manager->get() /srv/www/htdocs/owncloud/lib/private/User/Manager.php:139
8  OCA\User_LDAP\User\User->updateEmail() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:435
9  OCA\User_LDAP\User\User->update() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User/User.php:138
10 OCA\User_LDAP\User_LDAP->userExists() /srv/www/htdocs/owncloud/apps/user_ldap/lib/User_LDAP.php:277
[...]

PR #64, opened 2017-03-14
Correctly apply quota

jvillafanez:
Expected behaviour with these changes:
* If both LDAP quota field and default LDAP quota are set:
    * if the user has a value for the LDAP quota attribute, try to use that value as quota
    * if the user doesn't have a value for the LDAP quota attribute or the value can't be parsed correctly, use the default LDAP quota.
    * if the user doesn't have a value for the LDAP quota attribute or the value can't be parsed correctly, and the default LDAP quota can't be parsed correctly, use the ownCloud's default quota.
* If the LDAP quota field is empty and the default LDAP quota is set:
    * go through the above case and consider the user doesn't have the LDAP quota attribute set.
* If the LDAP quota field isn't empty and the LDAP default quota isn't set:
    * if the user has a value for the attribute, try to use that value as quota.
    * if the user doesn't have a value for the attribute **do not change anything**. This usually means that the ownCloud's default quota will be used. If the user's quota has a previous value, that value will be kept
* If neither the LDAP quota field nor the default LDAP quota is set, keep the current quota.

Additional changes:
* Explictly support "default" and "none" as valid values for the LDAP quota, both as default LDAP quota and as value for the attribute (not sure if it was supported before).
* Added new logs to troubleshoot weird quota strings.

Notes:
* Even if the user "uses" the default LDAP quota, that value **will be effectively set** as quota. If the admin remove the default LDAP quota at a later time, the previous default LDAP quota might still be used by other users.
* Changes in the quota aren't applied uniformly. They depend on when the user data is fetched from the LDAP. If you set a default LDAP quota value "A", then you change it to "B" and finally change it again to "C", quota values for different users might be "A", "B" or "C" at some point. It's expected that when you access to the user data (or the user logs in) the quota is applied correctly. However, keep in mind that there are some cases (check above) where the quota isn't changed, which will result in unexpected values for the quota.

@PVince81 @DeepDiver1975 to review
@owncloud/qa to verify and check the behaviour
@settermjd in case there is docs to be updated / improved due to this change

### DIFF
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -494,11 +494,16 @@ class User {
 			}
 		}
 
-		if($quota !== false) {
-			$this->userManager->get($this->uid)->setQuota($quota);
+		$targetUser = $this->userManager->get($this->uid);
+		if ($targetUser) {
+			if($quota !== false) {
+				$targetUser->setQuota($quota);
+			} else {
+				$this->log->log('not suitable default quota found for user ' . $this->uid . ': [' . $defaultQuota . ']', \OCP\Util::WARN);
+				$targetUser->setQuota('default');
+			}
 		} else {
-			$this->log->log('not suitable default quota found for user ' . $this->uid . ': [' . $defaultQuota . ']', \OCP\Util::WARN);
-			$this->userManager->get($this->uid)->setQuota('default');
+			$this->log->log('trying to set a quota for user ' . $this->uid . ' but the user is missing', \OCP\Util::ERROR);
 		}
 	}
 
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -455,6 +455,20 @@ class User {
 	}
 
 	/**
+	 * Overall process goes as follow:
+	 * 1. fetch the quota from LDAP and check if it's parseable with the "verifyQuotaValue" function
+	 * 2. if the value can't be fetched, is empty or not parseable, use the default LDAP quota
+	 * 3. if the default LDAP quota can't be parsed, use the ownCloud's default quota (use 'default')
+	 * 4. check if the target user exists and set the quota for the user.
+	 *
+	 * In order to improve performance and prevent an unwanted extra LDAP call, the $valueFromLDAP
+	 * parameter can be passed with the value of the attribute. This value will be considered as the
+	 * quota for the user coming from the LDAP server (step 1 of the process) It can be useful to
+	 * fetch all the user's attributes in one call and use the fetched values in this function.
+	 * The expected value for that parameter is a string describing the quota for the user. Valid
+	 * values are 'none' (unlimited), 'default' (the ownCloud's default quota), '1234' (quota in
+	 * bytes), '1234 MB' (quota in MB - check the \OC_Helper::computerFileSize method for more info)
+	 *
 	 * fetches the quota from LDAP and stores it as ownCloud user value
 	 * @param string $valueFromLDAP the quota attribute's value can be passed,
 	 * to save the readAttribute request
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -165,6 +165,10 @@ class User {
 		$attr = strtolower($this->connection->ldapQuotaAttribute);
 		if(isset($ldapEntry[$attr])) {
 			$this->updateQuota($ldapEntry[$attr][0]);
+		} else {
+			if ($this->connection->ldapQuotaDefault !== '') {
+				$this->updateQuota();
+			}
 		}
 		unset($attr);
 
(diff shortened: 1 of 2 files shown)


jvillafanez:
This should be enough for now. I plan to add some additional tests to cover as much cases as possible, going along with the current tests.

davitol:
:+1: WFM

> Changes in the quota aren't applied uniformly. They depend on when the user data is fetched from the LDAP. If you set a default quota value "A", then you change it to "B" and finally change it again to "C", quota values for different users might be "A", "B" or "C" at some point. It's expected that when you access to the user data (or the user logs in) the quota is applied correctly. However, keep in mind that there are some cases (check above) where the quota isn't changed, which will result in unexpected values for the quota.

As @jvillafanez said, it is needed to log in or show in the user's list the user in order to get the field configkey  set as 'quota' in the oc_preferences table.

davitol:
BTW I think we should change the default values from B to MB. (I know that it is possible to set XXX MB and works) but IMO byte is too small to be considered as Quota default unit. I'll open an enhancement ticket for that.

lib/User/User.php
```diff
@@ -460,23 +464,51 @@ public function updateQuota($valueFromLDAP = null) {
...
 
+		$quota = false;
 		if(is_null($valueFromLDAP)) {
 			$quotaAttribute = $this->connection->ldapQuotaAttribute;
 			if ($quotaAttribute !== '') {
 				$aQuota = $this->access->readAttribute($this->dn, $quotaAttribute);
 				if($aQuota && (count($aQuota) > 0)) {
-					$quota = $aQuota[0];
+					if ($this->verifyQuotaValue($aQuota[0])) {
+						$quota = $aQuota[0];
```
PVince81: I'm confused here even when trying to read the above code.

Can you add PHPDoc to remove all doubts ?

Is `$valueFromLDAP` the value of quota attribute from the specific user read from LDAP ?
If yes, then what is the difference between `$valueFromLDAP` and the `ldapQuotaAttribute` value that is read at this line ?

lib/User/User.php
```diff
@@ -460,23 +464,51 @@ public function updateQuota($valueFromLDAP = null) {
...
+					} else {
+						$this->log->log('not suitable LDAP quota found for user ' . $this->uid . ': [' . $aQuota[0] . ']', \OCP\Util::WARN);
+					}
 				}
 			}
+		} else {
+			if ($this->verifyQuotaValue($valueFromLDAP)) {
+				$quota = $valueFromLDAP;
```
PVince81: which value is this ? the default or the one read from the user's quot attribute ? (add PHPDoc to remove all doubts)

lib/User/User.php
```diff
@@ -460,23 +464,51 @@ public function updateQuota($valueFromLDAP = null) {
...
+
+		if ($quota === false) {
+			// quota not found using the LDAP attribute (or not parseable). Try the default quota
+			$defaultQuota = $this->connection->ldapQuotaDefault;
+			if ($this->verifyQuotaValue($defaultQuota)) {
+				$quota = $defaultQuota;
+			}
 		}
+
+		$targetUser = $this->userManager->get($this->uid);
```
PVince81: I assume that calling this here is not causing any recursion ? See https://github.com/owncloud/user_ldap/issues/47 for the reason why I'm asking

tests/User/UserTest.php
```diff
@@ -199,15 +199,17 @@ public function testUpdateQuotaAllProvided() {
...
+			->with($this->equalTo('ldapQuotaAttribute'))
+			->will($this->returnValue('myquota'));
 
+		/* Having a quota defined, the ldapQuotaDefault won't be used
 		$connection->expects($this->at(1))
 			->method('__get')
-			->with($this->equalTo('ldapQuotaAttribute'))
-			->will($this->returnValue('myquota'));
+			->with($this->equalTo('ldapQuotaDefault'))
+			->will($this->returnValue('23 GB'));
```
PVince81: did you comment out this test on purpose ?

tests/User/UserTest.php
```diff
@@ -289,15 +291,17 @@ public function testUpdateQuotaIndividualProvided() {
...
+			->with($this->equalTo('ldapQuotaAttribute'))
+			->will($this->returnValue('myquota'));
 
+		/* Having a quota set this won't be used
 		$connection->expects($this->at(1))
 			->method('__get')
-			->with($this->equalTo('ldapQuotaAttribute'))
-			->will($this->returnValue('myquota'));
+			->with($this->equalTo('ldapQuotaDefault'))
+			->will($this->returnValue(''));
```
PVince81: same question here

lib/User/User.php
PVince81: I see we did it before, but now we do it in a new code path. I guess it shouldn't hurt.

PVince81:
Missing test for the quota parser which now accepts "default" and "none" ?

PVince81:
Looks good in general. I think this is great for 10.0, not sure about backporting as this would change behaviors or potentially cause surprises if some admins were using the current behavior in a legitimate way, if that's even possible. @jvillafanez thoughts on this ?

cc @butonic

lib/User/User.php
jvillafanez: There might be a one step recursion cut down by the `wasRefreshed` check.

I don't think there is a good solution for this. Either we hack the `setQuota` call and bypass the user management to set the quota directly in the DB (as it was done in previous versions) or we fetch the wrapped user from the core user manager to set the quota. I don't see any other solution without changing the whole architecture.

tests/User/UserTest.php
jvillafanez: Yes, it was commented on purpose. Previously we were always fetching the default quota even though it wasn't being used in some scenarios. I think this way is a bit more clear that the default quota won't be fetched in this case
i'll check if I can do it properly.

jvillafanez: looks like I can't test that a method is never called with specific parameters... any other alternative?

lib/User/User.php
jvillafanez: PHPDoc updated with more information

lib/User/User.php
jvillafanez: PHPDoc updated

tests/User/UserTest.php
jvillafanez: It's commented on purpose, to make clear that the LDAP default quota won't be fetched in that case.
It doesn't seem possible to check if the method won't be called with the specified parameters, so I'm not sure if leave this commented or remove it.

jvillafanez:
I think the intention for the LDAP quota is more or less as explained in the PR, and everyone would expect the app to behave as explained (except for the corner cases, which I don't think we can solve). If the app isn't behaving as explained it should be considered as a bug.

There are already reports about strange quota behaviour, and this PR might solve them or at least help to debug what's happening with the quota.

PVince81:
Looks good, thanks 👍 

For the commented out test I don't have any ideas right now.

PVince81:
Please backport

jvillafanez:
Backport to stable9.1 in https://github.com/owncloud/core/pull/27420
Any additional backport needed?

PVince81:
If the behavior changed in OC 9 but was fine in OC 8.2 then we should likely also backport to stable9

jvillafanez:
Backport for stable9 in https://github.com/owncloud/core/pull/27424

IRAM38:
hello Everybody!
I have the same issue, I don't understand. I have Cloud Vers9.1.5 (stable)

Warning user_ldap not suitable default quota found for user xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

And the user Quota (for exemple 50Gb) is set to Default quota (4Gb) ..
Thanks for your help.

michaelstingl:
@IRAM38 No, I don't think you have the same issue. This is a [pull request](https://help.github.com/articles/about-pull-requests/) and the related issue has been fixed in 9.1.5: https://github.com/owncloud/core/pull/27420

https://central.owncloud.org/ is a good starting point for help.

PR #70, opened 2017-04-07
Trying to login the user before doing anything with it

SergioBertolinSG:
@jvillafanez @DeepDiver1975 I'm trying to get the user logged in before getting the avatar or displayname. Not working so far. Do you know what I am doing wrong?

### DIFF
--- a/tests/Integration/Lib/User/IntegrationTestUserAvatar.php
+++ b/tests/Integration/Lib/User/IntegrationTestUserAvatar.php
@@ -57,6 +57,8 @@ class IntegrationTestUserAvatar extends AbstractIntegrationTest {
 	 * @param string $image
 	 */
 	private function execFetchTest($dn, $username, $image) {
+		$uid = explode("=", explode(",", $dn)[0])[1];
+
 		$this->setJpegPhotoAttribute($dn, $image);
 
 		// assigns our self-picked oc username to the dn
@@ -72,12 +74,13 @@ class IntegrationTestUserAvatar extends AbstractIntegrationTest {
 		    $userObject->updateLastLoginTimestamp();
 		}
 		\OC_Util::setupFS($username);
+
+		$userSession = \OC::$server->getUserSession();
+		$userSession->login($uid, $uid);
+
 		if (\OC_User::userExists($username)) {
 		    \OC::$server->getUserFolder($username);
 		}
-
-		$userSession = \OC::$server->getUserSession();
-		$userSession->login($username, $username);
 
 		\OC::$server->getConfig()->deleteUserValue($username, 'user_ldap', User::USER_PREFKEY_LASTREFRESH);
 		if(\OC::$server->getAvatarManager()->getAvatar($username)->exists()) {
--- a/tests/Integration/Lib/User/IntegrationTestUserAvatar.php
+++ b/tests/Integration/Lib/User/IntegrationTestUserAvatar.php
@@ -69,24 +69,12 @@ class IntegrationTestUserAvatar extends AbstractIntegrationTest {
 		\OC_Util::tearDownFS();
 
 		$userSession = \OC::$server->getUserSession();
-		$userSession->login($uid, $uid);
-
-		\OC_User::setUserId($username);
-		$userObject = \OC::$server->getUserManager()->get($username);
-		if (!is_null($userObject)) {
-		    $userObject->updateLastLoginTimestamp();
+		if ($userSession->isLoggedIn()) {
+			$userSession->logout();
 		}
+		$userSession->login($username, $uid);
+
 		\OC_Util::setupFS($username);
-
-
-		if (\OC_User::userExists($username)) {
-		    \OC::$server->getUserFolder($username);
-		}
-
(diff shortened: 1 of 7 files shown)


PVince81:
no luck with core's TestCase `loginAsUser` method ?

SergioBertolinSG:
Some code comes from there: https://github.com/owncloud/core/blob/master/tests/lib/TestCase.php#L340-L348

I don't know if inheriting from that class is possible here and if that can make this work.

tests/Integration/Lib/User/IntegrationTestUserAvatar.php
```diff
@@ -65,8 +65,20 @@ private function execFetchTest($dn, $username, $image) {
...
 		\OC_Util::tearDownFS();
+
+		\OC_User::setUserId($username);
+		$userObject = \OC::$server->getUserManager()->get($username);
+		if (!is_null($userObject)) {
+		    $userObject->updateLastLoginTimestamp();
+		}
 		\OC_Util::setupFS($username);
-		\OC::$server->getUserFolder($username);
+		if (\OC_User::userExists($username)) {
```
jvillafanez: This will likely check the existence against the account table, but the user might not be there yet.

SergioBertolinSG: After moving 
```
$userSession = \OC::$server->getUserSession();
$userSession->login($username, $username);
```
before that check, nothing has changed. It is still failing at 
\OC::$server->getAvatarManager()->getAvatar($username)->exists()

PVince81:
xdebug stack trace for the new error:
```
0  OC\AvatarManager->getAvatar() /srv/www/htdocs/owncloud/lib/private/AvatarManager.php:86
1  OCA\User_LDAP\Tests\Integration\Lib\User\IntegrationTestUserAvatar->execFetchTest() /srv/www/htdocs/owncloud/apps3/user_ldap/tests/Integration/Lib/User/IntegrationTestUserAvatar.php:86
2  OCA\User_LDAP\Tests\Integration\Lib\User\IntegrationTestUserAvatar->case1() /srv/www/htdocs/owncloud/apps3/user_ldap/tests/Integration/Lib/User/IntegrationTestUserAvatar.php:105
3  OCA\User_LDAP\Tests\Integration\AbstractIntegrationTest->run() /srv/www/htdocs/owncloud/apps3/user_ldap/tests/Integration/AbstractIntegrationTest.php:127
4  {main}          /srv/www/htdocs/owncloud/apps3/user_ldap/tests/Integration/Lib/User/IntegrationTestUserAvatar.php:173

```

PVince81:
$user is null in AvatarManager

PVince81:
$userId is "alice1337".

So the user manager doesn't find this user

PVince81:
Ok, so the problem is simple: since this user hasn't logged in yet, there is no entry in the accounts table for that user. So when the AvatarManager is used by this test, it tries to get the user from the user manager which goes to the account table, but there is no entry there.

So you were right trying to simulate a login to make sure that account gets created.

PVince81:
I debugged into the login routine and while LDAP could get a $ldapRecord:
```
$ldapRecord                      = (array[4]);
  $ldapRecord['dn']              = (array[1]);
    $ldapRecord['dn'][0]         = (string[37]) 'uid=alice,ou=users,dc=owncloud,dc=com';
  $ldapRecord['uid']             = (array[1]);
    $ldapRecord['uid'][0]        = (string[5]) 'alice';
  $ldapRecord['displayname']     = (array[1]);
    $ldapRecord['displayname'][0] = (string[12]) 'Laeic, Alice';
```

the `$this->access->userManager->get($dn)` here returned null: https://github.com/owncloud/user_ldap/blob/fix_inttests/lib/User_LDAP.php#L127

PVince81:
the user does appear in oc_ldap_user_mapping:
```
MariaDB [owncloud]> select * from oc_ldap_user_mapping;
+---------------------------------------+---------------+--------------------+
| ldap_dn                               | owncloud_name | directory_uuid     |
+---------------------------------------+---------------+--------------------+
| uid=alice,ou=Users,dc=owncloud,dc=com | alice1337     | fakeUUID-alice1337 |
+---------------------------------------+---------------+--------------------+
```

Not sure what's wrong then... It looks like a LDAP config issue but since it worked in past versions, not sure how.

PVince81:
Tried adding `occ user:sync` in the script that runs before the test, but same issue.

My feeling is that the problem is on the LDAP side. Maybe the test doesn't setup properly.

@jvillafanez any clues ?

jvillafanez:
The login should be at the beginning of the test. There are some operations for the user manager that happen before the login:
```
		$userObject = \OC::$server->getUserManager()->get($username);
```

SergioBertolinSG:
@jvillafanez It fails also moving the login before that operations.

jvillafanez:
@SergioBertolinSG check if it works for you with the changes. Then we can merge into the fix_unittest branch and verify everything.

jvillafanez:
Note that the "isVisibleFor" change is already in master, but not yet into the fix_unittest branch. The change is required to make the tests work.
I expect the change won't be visible after merging this branch into the fix_unittest one. Anyway, after everything is in master, we should verify the change is there.

PR #76, opened 2017-04-19
Fix user import duplicated mails

jvillafanez:
Log a message and try to set the email as null if it's duplicated, logging again if it isn't possible. Duplicated emails will be ignored.


PVince81:
Ok as a quick fix, thanks 👍

Issue #83, opened 2017-04-26
Build script / Makefile for marketplace releasing


PVince81:
Easiest would be to reuse https://github.com/owncloud/market/blob/master/Makefile

@jvillafanez

DeepDiver1975:
I'm reopening this since the app cannot be uploaded to be marketplace:
- [ ] get a signing key
- [ ] add category to info.xml @pmaier1 where shall we put it?
- [ ] a screenshot has to be added to info.xml
- [ ] test it on marketplace.int.owncloud.com

pmaier1:
> add category to info.xml @pmaier1 where shall we put it?

'Integration' I would say.

pmaier1:
see [here](https://github.com/owncloud/marketplace/issues/85) in general ;)

DeepDiver1975:
:+1:

PVince81:
@jvillafanez can we finish this ? As I understand the LDAP app should be available in the market when 10.0.1 is out. Best would be to have it a bit earlier for testing (staging)

jvillafanez:
Makefile is already merged, so I guess this can be closed unless there are modifications to be made there.

PVince81:
@jvillafanez were these items handled ? https://github.com/owncloud/user_ldap/issues/83#issuecomment-299150776

PVince81:
@jvillafanez ping

jvillafanez:
I have my key and I can upload the package in the internal marketplace. I don't know if we should share the key so you can upload it to the internal marketplace if I'm out. Anyway, I don't expect to upload the app to the official marketplace.

jvillafanez:
Makefile ready and app uploaded. Closing.

PR #84, opened 2017-04-26
add script for marketplace releasing

jvillafanez:
Fix https://github.com/owncloud/user_ldap/issues/83

Note that there are updates for the dev libraries for phpunit. This shouldn't affect the app, just the tests, not sure how they will be affected though. We can use the old ones because all those dependencies are for phpunit.

**Please, verify all the data is correct**

@PVince81

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+	"directory": "vendor"
+}
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,101 @@
+SHELL := /bin/bash
+
+#
+# Define NPM and check if it is available on the system.
+#
+NPM := $(shell command -v npm 2> /dev/null)
+ifndef NPM
+    $(error npm is not available on your system, please install npm)
+endif
+
+PHPUNIT="$(PWD)/lib/composer/phpunit/phpunit/phpunit"
+
+ldap_doc_files=LICENSE README.md
+ldap_src_dirs=ajax appinfo css img js lib templates vendor
+ldap_all_src=$(ldap_src_dirs) $(ldap_doc_files)
+build_dir=build
+dist_dir=$(build_dir)/dist
+COMPOSER_BIN=$(build_dir)/composer.phar
+
+# internal aliases
+composer_deps=vendor/
+composer_dev_deps=lib/composer/phpunit
+js_deps=node_modules/
+
+#
+# Catch-all rules
+#
+.PHONY: all
+all: $(composer_dev_deps) $(js_deps)
+
+.PHONY: clean
+clean: clean-composer-deps clean-js-deps clean-dist clean-build
+
+
+#
+# Basic required tools
+#
+$(COMPOSER_BIN):
+	mkdir $(build_dir)
+	cd $(build_dir) && curl -sS https://getcomposer.org/installer | php
+
+#
+# ownCloud ldap PHP dependencies
+#
+$(composer_deps): $(COMPOSER_BIN) composer.json composer.lock
+	php $(COMPOSER_BIN) install --no-dev
+
+$(composer_dev_deps): $(COMPOSER_BIN) composer.json composer.lock
+	php $(COMPOSER_BIN) install --dev
+
+.PHONY: clean-composer-deps
+clean-composer-deps:
+	rm -f $(COMPOSER_BIN)
+	rm -Rf $(composer_deps)
+
+.PHONY: update-composer
+update-composer: $(COMPOSER_BIN)
+	rm -f composer.lock
+	php $(COMPOSER_BIN) install --prefer-dist
+
+#
+# ownCloud ldap JavaScript dependencies
+#
+$(js_deps): $(NPM) package.json
+	$(NPM) install
+	touch $(js_deps)
+
+.PHONY: install-js-deps
+install-js-deps: $(js_deps)
+
+.PHONY: update-js-deps
+update-js-deps: $(js_deps)
+
+
+.PHONY: clean-js-deps
+clean-js-deps:
+	rm -Rf $(js_deps)
+
+#
+# dist
+#
+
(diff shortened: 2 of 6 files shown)


DeepDiver1975:
signing is missing - see https://github.com/owncloud/contacts/blob/master/Makefile#L53

jvillafanez:
Signing added but untested. The code goes through the fallback route and generate a tarball.

DeepDiver1975:
So.... get a  key ....

PVince81:
@jvillafanez please retest the LDAP app after running `make` and also the one produced by `make dist` by copying it to a tarball release.

The local `make` breaks it:
ldap-borked

jvillafanez:
It should be fixed now, although I don't like how it is done 😞 

There are a couple of things that I have changed in my personal environment to test it:

* The NODE_PREFIX is set to the same value as the one from the customgroups Makefile https://github.com/owncloud/customgroups/blob/master/rules/deps.mk

    I had to change it to use the one from the core

* I was using a docker environment with a root account, so all the bower commands were changed to use the --allow-root option. This shouldn't needed (that option isn't committed)

We should move the js dependencies (and a cleanup of the makefile) when we have time

PVince81:
```
~/work/workspace/owncloud/apps3/user_ldap [makefile_for_market *]
± % make
mkdir /srv/www/htdocs/owncloud/apps3/user_ldap/build
cd /srv/www/htdocs/owncloud/apps3/user_ldap/build && curl -sS https://getcomposer.org/installer | php
All settings correct for using Composer
Downloading...

Composer (version 1.4.1) successfully installed to: /srv/www/htdocs/owncloud/apps3/user_ldap/build/composer.phar
Use it: php composer.phar

php /srv/www/htdocs/owncloud/apps3/user_ldap/build/composer.phar install --dev
You are using the deprecated option "dev". Dev packages are installed by default now.
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Package operations: 25 installs, 0 updates, 0 removals
  - Installing webmozart/assert (1.2.0): Loading from cache
  - Installing phpdocumentor/reflection-common (1.0): Loading from cache
  - Installing phpdocumentor/type-resolver (0.2.1): Loading from cache
  - Installing phpdocumentor/reflection-docblock (3.1.1): Loading from cache
  - Installing phpunit/php-token-stream (1.4.11): Loading from cache
  - Installing symfony/yaml (v3.2.7): Loading from cache
  - Installing sebastian/version (2.0.1): Loading from cache
  - Installing sebastian/resource-operations (1.0.0): Loading from cache
  - Installing sebastian/recursion-context (2.0.0): Loading from cache
  - Installing sebastian/object-enumerator (2.0.1): Loading from cache
  - Installing sebastian/global-state (1.1.1): Loading from cache
  - Installing sebastian/exporter (2.0.0): Loading from cache
  - Installing sebastian/environment (2.0.0): Loading from cache
  - Installing sebastian/diff (1.4.1): Loading from cache
  - Installing sebastian/comparator (1.2.4): Loading from cache
  - Installing phpunit/php-text-template (1.2.1): Loading from cache
  - Installing doctrine/instantiator (1.0.5): Loading from cache
  - Installing phpunit/phpunit-mock-objects (3.4.3): Loading from cache
  - Installing phpunit/php-timer (1.0.9): Loading from cache
  - Installing phpunit/php-file-iterator (1.4.2): Loading from cache
  - Installing sebastian/code-unit-reverse-lookup (1.0.1): Loading from cache
  - Installing phpunit/php-code-coverage (4.0.8): Loading from cache
  - Installing phpspec/prophecy (v1.7.0): Loading from cache
  - Installing myclabs/deep-copy (1.6.1): Loading from cache
  - Installing phpunit/phpunit (5.7.19): Loading from cache
symfony/yaml suggests installing symfony/console (For validating YAML files using the lint command)
sebastian/global-state suggests installing ext-uopz (*)
phpunit/phpunit-mock-objects suggests installing ext-soap (*)
phpunit/phpunit suggests installing phpunit/php-invoker (~1.1)
Generating autoload files
/home/vincent/work/workspace/owncloud/apps3/user_ldap/node_modules/bower/bin/bower install && touch vendor/
/bin/bash: /home/vincent/work/workspace/owncloud/apps3/user_ldap/node_modules/bower/bin/bower: No such file or directory
Makefile:68: recipe for target 'lib/composer/phpunit' failed
make: *** [lib/composer/phpunit] Error 127
```

Bower doesn't seem to be found. Maybe npm didn't run yet ?

@jvillafanez when I asked for a Makefile I didn't mean to move all dependencies to composer, npm, bower. Just have a simple makefile that properly builds the tarball and strips it from unwanted files is enough as a first step.

Making bower download whatever deps the app used to bundle/require can be a time consuming process...

jvillafanez:
The problem is that the js dependencies are in the "vendor" folder, which is also used by composer. Doing a make clean would remove the composer dependencies along with the bundled js library.

Work for me now just with the "--allow-root" fix. If it doesn't work I'll provide the ugliest handcrafted Makefile ever.

PVince81:
@jvillafanez you could move JS deps in "js/vendor".

I wonder how this can even work if JS deps are inside "vendor", because as far as I know the resource locator for JS `addScript()` only ever looks inside of the "js" folder. Maybe the JS deps are unused ?

jvillafanez:
"vendor_script" and "vendor_style" are being used to load those from the vendor folder.

I don't think it's possible to move the dependencies because the library requires some custom css that needs to be loaded (included with the library). I've tried to move it, but the css needed will be saved in the js/vendor folder and we can't load the css from there easily.
Adding additional logic to move the css to another folder doesn't seem a good idea.

PVince81:
ah ok, I didn't know it was possible to load JS from outside of "js".

PVince81:
Works now 👍 

Tested by running `make` then setting up LDAP, syncing users and checking users page. The users appear.

PR #85, opened 2017-04-27
Remove offline user and remnants table (step1)

jvillafanez:
First steps for a better integration.

TODO things:
* [x] review and remove old code that might be lurking around about configuration values, user preferences, and other stuff (currently done a quick pass, but there might be more things to clean)
* [x] adjust tests

## NOTES
There are several behavioural changes with this PR:
* LDAP accounts can be removed from the web UI. This implies the ownCloud account will be removed, the LDAP mapping will also be removed, and the data stored in ownCloud will be removed. **The user's information in the LDAP server will remain untouched**
* Even if this all the data stored in ownCloud is removed, the LDAP user can be synced again as long as he is in the LDAP server. *As expected* all his previous data will be lost.

In order to delete an user for the whole system (with this PR), the recommended way would be:
1. Remove the user from the LDAP server (obviously, you'll remove the user from other systems that are using that LDAP server.
2. You can verify the user isn't in the LDAP server anymore with the `./occ ldap:check-user` command. This might not be mandatory but recommended in order to verify the user isn't in the system.
3. Use the web UI or the `./occ user:delete` to delete the user

As I've said, you can remove the data stored in ownCloud by deleting the user, but that user can still log in and be synced, although his data won't be there.


jvillafanez:
In order to prevent user from login without deleting their data...
* `occ user:disable` works fine. No problems found
* Modify the ldap login filter to exclude specific users. Usually adding a flag in the filter is enough, for example `(&(objectClass=person)(uid=%uid)(status=active))` (note that the "status" attribute in the example won't likely exist and you likely need to use another)

Both methods seem to work fine. You need both "checks" pass in order to log in successfully.

Not sure if there are other methods to be checked.

jvillafanez:
Removed old information in LDAP which shouldn't be used any longer (it has been moved to the accounts table). The only two keys remaining are "firstLoginAccomplished" and "lastFeatureRefresh", to be evaluated what could happen if those are removed.

The new commit doesn't seem to introduce any further change in the behavior

davitol:
👍 No issues were found until now so I feel OK about merging this from QA POV

jvillafanez:
Please check https://github.com/owncloud/user_ldap/pull/85/commits/fbd044a91f006e3da728f51671440cd3cfaa861b for the changes in the tests

DeepDiver1975:
okay - let's merge this then ....

PR #99, opened 2017-05-22
Increase minimum version to 10.0.1

jvillafanez:
For marketplace release.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -20,7 +20,7 @@ A user logs into ownCloud with their LDAP or AD credentials, and is granted acce
 	<screenshot>https://raw.githubusercontent.com/owncloud/screenshots/0527af6b981603f3410982c2ed70051b7cf991c3/user_ldap/user_ldap.png</screenshot>
 	<dependencies>
 		<lib>ldap</lib>
-		<owncloud min-version="10.0" max-version="10.0" />
+		<owncloud min-version="10.0.1" max-version="10.0.99" />
 	</dependencies>
 
 	<namespace>User_LDAP</namespace>


jvillafanez:
@PVince81

appinfo/info.xml
```diff
@@ -20,7 +20,7 @@ A user logs into ownCloud with their LDAP or AD credentials, and is granted acce
 	<screenshot>https://raw.githubusercontent.com/owncloud/screenshots/0527af6b981603f3410982c2ed70051b7cf991c3/user_ldap/user_ldap.png</screenshot>
 	<dependencies>
 		<lib>ldap</lib>
-		<owncloud min-version="10.0" max-version="10.0" />
+		<owncloud min-version="10.0.1" max-version="10.0.99" />
```
DeepDiver1975: may can stay at 10.0 - this will be interpreted as 10.0.9999 on the marketplace

jvillafanez: I don't know how it will be interpreted when ownCloud tries to install it. From my undestanding, 10.0.1 > 10.0 so the min version will be higher than the max version, at least in ownCloud.
I haven't tested that, but feel free to change it.

PVince81:
@jvillafanez missing statement about how / whether you tested this

butonic:
👍

jvillafanez:
Tested steps:
1. Generate app package with this PR change
2. Upload app package to internal marketplace
3. Make sure this is the only version uploaded
4. Install OC 10.0.0
5. Use `occ market:list` and `occ market:install user_ldap`

The app isn't listed and trying to install will return an "unknown app" error

PR #100, opened 2017-05-22
Always set a proper prefix on CLI

tboerger:
I had to remove the clause to being able to always set values via the
CLI. With this condition you can not configure the first group on the
CLI, without that condition you always get `s01` as a first prefix and
you are able to configure it on the CLI

Signed-off-by: Thomas Boerger <tboerger@owncloud.com>

### DIFF
--- a/lib/Command/CreateEmptyConfig.php
+++ b/lib/Command/CreateEmptyConfig.php
@@ -60,11 +60,6 @@ class CreateEmptyConfig extends Command {
 	protected function getNewConfigurationPrefix() {
 		$serverConnections = $this->helper->getServerConfigurationPrefixes();
 
-		// first connection uses no prefix
-		if(sizeof($serverConnections) == 0) {
-			return '';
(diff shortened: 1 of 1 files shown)


felixboehm:
- [ ] Change docs, here https://doc.owncloud.com/server/10.0/admin_manual/configuration_server/occ_command.html?highlight=command#ldap-commands

```
ldap:create-empty-config creates an empty LDAP configuration. The first one you create has no configID, like this example:

sudo -u www-data php occ ldap:create-empty-config
  Created new configuration with configID ''
This is a holdover from the early days, when there was no option to create additional configurations. The second, and all subsequent, configurations that you create are automatically assigned IDs:

sudo -u www-data php occ ldap:create-empty-config
   Created new configuration with configID 's01'
```

PVince81:
documentation is best reported in a documentation ticket in the repo

cc  @settermjd

PR #101, opened 2017-05-22
allow proxy to check next server

butonic:
fix getting search terms in case the different ldap servers use different base dns.

no release blocker. can be released via market. cc @DeepDiver1975 for releasing asap?


cdamken:
👍  Tested and Works!

PR #105, opened 2017-06-05
Add hint for max search term length

tomneedham:
https://github.com/owncloud/core/pull/28019


jvillafanez:
What is limited is the value of the attribute, not the attribute.

What could happen if the length is greater that 191 chars? Taking into account admins won't have full control over it, they should know the risk.

tomneedham:
> What could happen if the length is greater that 191 chars? Taking into account admins won't have full control over it, they should know the risk.

truncated

jvillafanez:
👍

PR #109, opened 2017-06-26
remove unneeded account updates

butonic:
While working on #28212 I was profiling the performance for user:sync and this actually made a measurable difference. The user:sync command already has explicit update logic. The update logic in checkPassword remains.


butonic:
- [ ] add interface for avatar provisioning

tomneedham:
tested at customer

tomneedham:
rebased

PR #116, opened 2017-07-19
Add groupOfUniqueNames to the group list

pboyd04:
If you use LDAP with group OfUniqueNames instead of groupOfNames for group lists they don't show up in the wizard. This fixes that and makes it much easier for a user to setup in this situation.


tomneedham:
Hey @pboyd04 thanks for the PR. Do you mind rebasing this? And @jvillafanez could you review this? Thanks :)

jvillafanez:
@pboyd04 in the `determineGroupObjectClasses` function you should probably add the item as well. I think that's the only thing missing. For the rest, it's fine.

pboyd04:
That function already has * in the object class list. So I'm not sure it really needs it.

jvillafanez:
I'm not sure if it's required, but it should be there.

pboyd04:
Ok, done.

PR #119, opened 2017-08-03
Fix handle request to correctly call backend when cache is empty

tomneedham:
This allows the handleRequest method to call the userBackend, even when the caches return empty values. Without this, callOnLastSeenOn() was returning false, which was compared to $passOnWhen - whereas, it should pass through $passOnWhen for the comparison.

This fixes an issue where the search terms are not correctly saved during user sync

### DIFF
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -118,7 +118,7 @@ class User_Proxy extends Proxy implements IUserBackend, UserInterface, IProvides
 				return $result;
 			}
 		}
-		return false;
+		return $passOnWhen;
 	}
 
 	/**


jvillafanez:
We need to adjust the PHPDoc. Other than that 👍

butonic:
obsoleted by https://github.com/owncloud/user_ldap/pull/125

PVince81:
but https://github.com/owncloud/user_ldap/pull/125 is not ready and this PR here is very small

Merge anyway ?

tomneedham:
Will #125 make 10.0.3?

PVince81:
@tom irrelevant as user_ldap can be released separately, unless there are core changes required...

tomneedham:
Will address in #125 tomorrow

PR #120, opened 2017-08-04
user_ldap: Makes the time between needsRefresh configurable

sidey79:
**Description**

Makes the time between needsRefresh configurable via app config option updateAttribuesInterval.
Default is still 86400 secons which is one day.

**Motivation and Context**

Modify the interval without changing the sourcecode to make the interval persistent when updating

**Types of changes**

- ~~Bug fix (non-breaking change which fixes an issue)~~
- [x] New feature (non-breaking change which adds functionality)
- ~~Breaking change (fix or feature that would cause existing functionality to change)~~

**Checklist:**

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x]  updated the documentation accordingly.  ->(https://github.com/owncloud/documentation/issues/3278#issuecomment-321467584)
- [x]  I have read the CONTRIBUTING document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -357,8 +357,7 @@ class User {
 		$lastChecked = $this->config->getUserValue($this->uid, 'user_ldap',
 			self::USER_PREFKEY_LASTREFRESH, 0);
 
-		//TODO make interval configurable
-		if((time() - intval($lastChecked)) < 86400 ) {
+		if((time() - intval($lastChecked)) < $this->config->getAppValue('user_ldap', 'updateAttribuesInterval', 86400) ) {
 			return false;
 		}
 		return  true;
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -357,7 +357,7 @@ class User {
 		$lastChecked = $this->config->getUserValue($this->uid, 'user_ldap',
 			self::USER_PREFKEY_LASTREFRESH, 0);
 
-		if((time() - intval($lastChecked)) < $this->config->getAppValue('user_ldap', 'updateAttribuesInterval', 86400) ) {
+		if((time() - intval($lastChecked)) < $this->config->getAppValue('user_ldap', 'updateAttributesInterval', 86400) ) {
 			return false;
 		}
 		return  true;
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -357,7 +357,7 @@ class User {
 		$lastChecked = $this->config->getUserValue($this->uid, 'user_ldap',
 			self::USER_PREFKEY_LASTREFRESH, 0);
 
-		if((time() - intval($lastChecked)) < $this->config->getAppValue('user_ldap', 'updateAttributesInterval', 86400) ) {
+		if((time() - intval($lastChecked)) < intval($this->config->getAppValue('user_ldap', 'updateAttributesInterval', 86400)) ) {
 			return false;
 		}
 		return  true;
--- a/tests/User/UserTest.php
+++ b/tests/User/UserTest.php
@@ -1177,8 +1177,15 @@ class UserTest extends \Test\TestCase {
 			->with($this->equalTo('alice'), $this->equalTo('user_ldap'),
 				$this->equalTo(User::USER_PREFKEY_LASTREFRESH),
 				$this->equalTo(0))
-			->will($this->returnValue(time()));
-
+				->will($this->returnValue(time() - 10));
+		
+		$config->expects($this->once())
+			->method('getAppValue')
+			->with($this->equalTo('user_ldap'),
+				$this->equalTo('updateAttributesInterval'),
+				$this->anything())
+			->will($this->returnValue(1800);
+			       
 		$config->expects($this->exactly(2))
 			->method('getUserValue');
 
--- a/tests/User/UserTest.php
+++ b/tests/User/UserTest.php
@@ -1177,7 +1177,7 @@ class UserTest extends \Test\TestCase {
 			->with($this->equalTo('alice'), $this->equalTo('user_ldap'),
 				$this->equalTo(User::USER_PREFKEY_LASTREFRESH),
 				$this->equalTo(0))
-				->will($this->returnValue(time() - 10));
+			->will($this->returnValue(time() - 10));
 		
 		$config->expects($this->once())
 			->method('getAppValue')
--- a/README.md
+++ b/README.md
@@ -16,3 +16,11 @@
     * This option affects all LDAP connections. It isn't possible to enable this option for a specific connection.
     * This option could have a performance impact on big LDAP installations. Check your LDAP provider how to enable indexes for medial searches if they're supported but not active.
     * The option will work regardless of whether the LDAP server has an index for this. Small LDAP installations could have an acceptable performance with this option active even if the LDAP doesn't have that index active.
+
+#### Additional configuration options that can be modified via occ configurations 
+
+The user_ldap app refreshs will check for updated attributes at every user login. Attributes like mail oder quota are retrieved from the ldap server. To save ressources on the ldap server, there is a minimum time between two updates for every user. Without modification, the update intervall is not more often then 86400 seconds (1 day). This can be modifed by setting the app config 'updateAttributesInterval' to any number you like. Setting this value to 0 will update on very login request which can be quiet often and stress your ldap server.
+Attributes are unlikely to change very often, but waiting a day for a new quota is maybe a little bit long.
+
+To allow modify the minimum time between checks to one hour (3600 seconds) you can do this via occ:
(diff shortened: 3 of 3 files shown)


jvillafanez:
@sidey79 could you provide something in order to let the admins set that value? The change makes sense, but we can't ask admins to modify the DB to set the value.

PVince81:
@jvillafanez they could use the occ command as an alternative. In this case best have this setting in the documentation.

jvillafanez:
Ok, should we open a ticket in the documentation repo or update the Readme.md file? maybe both?

PVince81:
@jvillafanez up to you

sidey79:
@jvillafanez
@PVince81

Modify tue value via occ was what i was thinking about. Today wenn have to change the sourcecode which is not really nice.

So i have to open a new pr for the documentation, because it is on another repo?

lib/User/User.php
```diff
@@ -357,8 +357,7 @@ private function needsRefresh() {
 		$lastChecked = $this->config->getUserValue($this->uid, 'user_ldap',
 			self::USER_PREFKEY_LASTREFRESH, 0);
 
-		//TODO make interval configurable
-		if((time() - intval($lastChecked)) < 86400 ) {
+		if((time() - intval($lastChecked)) < $this->config->getAppValue('user_ldap', 'updateAttribuesInterval', 86400) ) {
```
jvillafanez: Fix typo: updateAttributesInterval

jvillafanez:
I've already opened a ticket in the documentation repo. Short term, check my comment above and update the Readme.md file to include that option, just a couple of lines explaining that the option is available and how you can set the value. This way new user can track new options available.

In addition, could you check why unittests are failing? https://travis-ci.org/owncloud/user_ldap/jobs/261160057 Ping me if you need help with this.

(in case you can't access):
```
$ if [ $(phpenv version-name) != "7.1" ]; then vendor/bin/phpunit ; else vendor/bin/phpunit --no-coverage ; fi

PHPUnit 5.7.15 by Sebastian Bergmann and contributors.
Runtime:       PHP 5.6.24 with Xdebug 2.4.1
Configuration: /home/travis/build/owncloud/core/apps/user_ldap/phpunit.xml

...............................................................  63 / 149 ( 42%)
...................................F........................... 126 / 149 ( 84%)
.......................                                         149 / 149 (100%)

Time: 43.86 seconds, Memory: 52.00MB

There was 1 failure:

1) OCA\User_LDAP\Tests\User\UserTest::testUpdateNoRefresh
OCP\IConfig::setUserValue('alice', 'user_ldap', 'lastFeatureRefresh', 1501880305, null) was not expected to be called.
/home/travis/build/owncloud/core/apps/user_ldap/lib/User/User.php:347
/home/travis/build/owncloud/core/apps/user_ldap/lib/User/User.php:154
/home/travis/build/owncloud/core/apps/user_ldap/tests/User/UserTest.php:1194
/home/travis/build/owncloud/core/apps/user_ldap/vendor/phpunit/phpunit/src/TextUI/Command.php:186
/home/travis/build/owncloud/core/apps/user_ldap/vendor/phpunit/phpunit/src/TextUI/Command.php:116

FAILURES!
Tests: 149, Assertions: 380, Failures: 1.
```

sidey79:
@jvillafanez   I've access to the travis ci log, no problem at all with this.

I've fixed the typo and addes a description to the readme.
The failing unit tests provide me some headaches.  Not totaly clear to me why it's failing. I've no idea how to set up a a clome thom the unit test environment on my own system. 
I think, the unittest problem my be caused, because the value is not interpreded as integer. Fixed this and now i hope for a new build check.

jvillafanez:
You'll need to adjust the tests. The problem is that, in the tests, the configuration object is mocked. Once the execution reaches your new code, your method will return `null` because the mocked object doesn't have any value defined for your method.

As said, you'll need to adjust the failing test.
First, adjust the expected `$lastChecked` variable:
```
		$config->expects($this->at(1))
			->method('getUserValue')
			->with($this->equalTo('alice'), $this->equalTo('user_ldap'),
				$this->equalTo(User::USER_PREFKEY_LASTREFRESH),
				$this->equalTo(0))
			->will($this->returnValue(time() - 10));
```
this means that `$this->config->getUserValue($this->uid, 'user_ldap',self::USER_PREFKEY_LASTREFRESH, 0)` will return a timestamp 10 seconds before the current one. When the execution reaches your code, the `if((time() - intval($lastChecked)) < 86400 ) {` will be something like `time() - (time() - 10) < 86400` (change the 86400 for whatever the value is returned by the configuration). Right now, the value returned by the configuration object is `null`, so basically that check is currently failing.

In addition, you'll need to make sure the app value within those 10 seconds.
```
		$config->expects($this->once())
			->method('getAppValue')
			->with($this->equalTo('user_ldap'),
				$this->equalTo('updateAttributesInterval'),
				$this->anything())
			->will($this->returnValue(5);
```
Now, when the test reaches your code, the configuration will return `5` when asked for the `getAppValue('user_ldap', 'updateAttributesInterval', 86400)`. The condition will fail and the execution should go through the previous way.

That should fix the tests.

sidey79:
@jvillafanez
Thanksgiving dir the example. Now i have a netter feeling about the unittests.

So the root for the failing test is, that the config object ist mit really the same as in an owncloud installation.
This causes that the method getAppValue is not  returning the default value, i've provided.

So if i modify the tests as proposed, the tests will simulate that 'updateAttributesInterval' ist set to 5. 

To test the default behavior another test would be needed. Can i expected that the  correct working of getAppValue ist covered in a tests in the core repository and i don't need to care about this?

jvillafanez:
> To test the default behavior another test would be needed. Can i expected that the correct working of getAppValue ist covered in a tests in the core repository and i don't need to care about this?

Officially, you should also check some other tests that are passing because, as said, the method would be returning `null` on those cases, but I can take care of those once this is merged. There shouldn't be needed to add more tests because all the possible paths should be covered already.

Just focus on fixing the test.

I'd recommend you to read about unittesting, in particular phpunit: https://phpunit.de/manual/current/en/index.html

sidey79:
@jvillafanez

I hope i got it right now. Changed the return value for updateAttributesInterval to 1800, because i think it needs to be higher than 10.

But i'm not able to run the unit tests my own. Sorry for that, but i have not found any docs on how to set up a working testing environment.

sidey79:
Think i got a running unittest and fixed the tests now:

```
Runtime:       PHP 7.0.7 with Xdebug 2.4.0
Configuration: /home/travis/build/sidey79/core/apps/user_ldap/phpunit.xml
...............................................................  63 / 149 ( 42%)
............................................................... 126 / 149 ( 84%)
.......................                                         149 / 149 (100%)
Time: 15.47 seconds, Memory: 36.00MB
OK (149 tests, 381 assertions)
```

tests/User/UserTest.php
```diff
@@ -1177,8 +1177,15 @@ public function testUpdateNoRefresh() {
 			->with($this->equalTo('alice'), $this->equalTo('user_ldap'),
 				$this->equalTo(User::USER_PREFKEY_LASTREFRESH),
 				$this->equalTo(0))
-			->will($this->returnValue(time()));
-
+				->will($this->returnValue(time() - 10));
```
jvillafanez: Adjust indentation here.

jvillafanez:
Ok, just that little thing I've commented above and we're ready to go.

sidey79:
@jvillafanez 
Okay, i've got your point. Thank you for this great support. Now i know a little bit mor about mocking objects. 👍 

Do you think this change can get released in alignment to 10.0.3?

jvillafanez:
This app has been already moved to the marketplace and has its own lifecycle. This means that these changes will be available in future versions of the ldap app, which isn't attached any longer to the ownCloud's core, so it could be available for the current 10.0.2 core version.

I'm not in charge of apps releasing, so I don't know when a new version of the app will be available in the marketplace. However, once this is merged, the next version of the app should have these changes.

sidey79:
@jvillafanez 

I did some tests on my own and i am not absolutly sure what is the right behaviour:


`occ config:app:set user_ldap updateAttributesInterval  --value=3600 --type=integer`
does not work, because --type is not known by occ command. I read the docs and it seems that they are only available  in `config:system:set` .

Is that right?

jvillafanez:
`sudo -u www-data ./occ config:app:set -h` There isn't any `--type` option for that command.

sidey79:
Docs have been updated, should be merged i think

sidey79:
@jvillafanez

Is there anything open from my side to merge this pull request?
The docs Repo was updated, but the app hasn't the feature.

Issue #111, opened 2017-06-28
Warning user_ldap not suitable default quota


IRAM38:
hello Everybody!
I have the same issue, I don't understand. I have Cloud Vers9.1.5 (stable)

Warning user_ldap not suitable default quota found for user xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

And the user Quota (for exemple 50Gb) is set automatically to Default quota (4Gb) ..

Thanks for your help.

cristiy2007:
Hello,

I have the exact same problem, I set the quota at 50 GB, and after 1 days aprox, it gets set to the default quota for 1 GB. I have Owncloud 9.1.5.
I will be updating today to 9.1.6 but I don't see anuthing related to this issue in the changelog so I don't expect it will be fixed.
Users quota is only defined in Owncloud, nothing from ldap.

Thanks

PVince81:
please first upgrade to 9.1.6, see changelog item:

"[major] Default quota display in users page now works correctly again - https://github.com/owncloud/core/issues/27896"

If the problem persists then it could be https://github.com/owncloud/core/issues/28380

nfc036:
We observed this issue as well and from my investigation it is a problem of user_ldap. [Here](https://github.com/owncloud/user_ldap/blob/master/lib/User/User.php#L553) the module is setting the quota to default if there are no relevant Attributes in LDAP. The function should take into account the already existing entries. Changing the given line to `$targetUser->setQuota($targetUser->getQuota())` should keep the existing entry or set a new entry 'default' ('default' is the standard return value of getQuota).

nfc036:
pull request #122 created.

cristiy2007:
Hello,

Please note that I completed the update to Owncloud 9.1.6 and I am still having the same issue. It happens random, take s betwen 24-48 hours and it affects random users, when it does I see this message in the logs:
```"...app":"user_ldap","message":"not suitable default quota found for user user1: []","level":2,"time":"2017-08-21T05:09:58+00:00","method":"PROPFIND","url":"\/remote.php\/webdav\/","user":"user2"}
"```

and it actually seems to happen when someonelse is trying to access/modify a share file or folder, so when user2 is accessing files/folders shared from user1 it gives the error "not suitable quota found for user user1" and resets his quota.

We're using this for 500 users so it's hard to keep track every morning to check which quota got changed to the default instead of the used configured.

PR #122, opened 2017-08-15
Do not overwrite quota entries

nfc036:
If a user already has a quota, this value should not be changed to 'default' on a daily basis...

### DIFF
--- a/lib/User/User.php
+++ b/lib/User/User.php
@@ -550,7 +550,7 @@ class User {
 				$targetUser->setQuota($quota);
 			} else {
 				$this->log->log('not suitable default quota found for user ' . $this->uid . ': [' . $defaultQuota . ']', \OCP\Util::WARN);
-				$targetUser->setQuota('default');
+				$targetUser->setQuota($targetUser->getQuota());
 			}
 		} else {
 			$this->log->log('trying to set a quota for user ' . $this->uid . ' but the user is missing', \OCP\Util::ERROR);


nfc036:
fixes issue #111

nfc036:
Tests failed, because the mock for IUser is not returning 'default' in getQuota(). As I am not familiar with this project, I am unable to correct the test code.

PVince81:
@jvillafanez "no suitable quota found", isn't that the code path that is never supposed to happen ?

I don't think this fix is correct but if it does work, it might give more clues about the original issue.
The question is why the quota value itself cannot be read in the first place.

nfc036:
user_ldap is not calling getQuota(), but is (currently) updating the quota for every (active) user once a day. There might be better ways to fix this, but the issue is currently a problem for our users/owncloud administrators. Any fix is highly appreciated.

PVince81:
If I understand correctly, you didn't setup any LDAP attributes for quota ? Or the LDAP attribute quota value is empty. In this use case, your PR makes sense, the code should not attempt to set the quota to "default" if there is already a value there.

@jvillafanez please review

nfc036:
Correct, we do not use any LDAP attributes for quotas.

jvillafanez:
We've opted to remove that line for stable 9.1 (still to be merged) https://github.com/owncloud/core/pull/28404 We'll need to port the solution.

PVince81:
Closing in favor of https://github.com/owncloud/core/pull/28404 which existed before.

@nfc036 still, thanks looking into this, at least we know what the solution is to this specific problem.

@jvillafanez let's move forward with the ports

PR #130, opened 2017-09-27
First go at migration of empty ldap configs

tomneedham:
Update migration to move empty config ids to something not empty.

Now to see what breaks.

Should I flush out the old cache?

I need to write some tests for this.


tomneedham:
need to make sure we always use a prefix - i  think we did this somewhere else already...

tomneedham:
Steps to test:
 - setup owncloud 9 with user_ldap
 - add two configurations (one will have an empty config prefix)
 - upgrade to 10 using this patch in the ldap app
 - check that the configurations have been migrated and the users from both configurations can login

tomneedham:
SQL to change a ldap prefix for testing

```sql
UPDATE oc_appconfig SET configkey = CONCAT('s07', SUBSTRING(configkey, 4, LENGTH(configkey))) WHERE appid='user_ldap' AND configkey LIKE 's04%' ;
```

felixboehm:
nice!

tomneedham:
Looks like php7 is now compiled with ldap in travis so we dont need to include it https://github.com/travis-ci/travis-ci/issues/7067#issuecomment-329145220

lib/Migrations/Version20170927125822.php
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	/**
+     * @param IOutput $out
+     */
+    public function run(IOutput $out) {
+        // Find out what configurations we are using and modify them
+
+		$configPrefixes = $this->helper->getServerConfigurationPrefixes();
+		foreach($configPrefixes as $prefix) {
```
tomneedham: I should just find the empty one

lib/Migrations/Version20170927125822.php
```diff
@@ -0,0 +1,46 @@
...
+				// Migrate the config
+				$newConfig = new Configuration($newPrefix, false);
+				foreach($oldConfig->getDefaults() as $configKey) {
+					// Set the new config equal to the old config, or default
+					$newConfig->$configKey = $oldConfig->$configKey;
+				}
+				$newConfig->saveConfiguration();
+				$this->helper->deleteServerConfiguration('');
+			}
+		}
```
tomneedham: Before the migration - the app is loaded by the MigrationService, which will initialise the User backend using the old prefix. So after the migration we need to before this configuration from the user proxy and insert a new one - otherwise things will start to 💥

tomneedham: Could add a method to User_Proxy to remove a backend given a configuration...

tomneedham: Cache will be invalidated because the configuration prefix is used in the cache key https://github.com/owncloud/user_ldap/blob/9e3117d6a03136d14cc4aa409d7bed4f342feb21/lib/Connection.php#L200 - could lead to longer migrations / upgrades after this migration is executed

tomneedham: If we tie this PR to #125 then caching is not an issue - unless many groups are present

lib/Migrations/Version20170927125822.php
```diff
@@ -0,0 +1,46 @@
...
+
+	public function __construct(Helper $helper) {
+		$this->helper = $helper;
+	}
+
+	/**
+     * @param IOutput $out
+     */
+    public function run(IOutput $out) {
+        // Find out what configurations we are using and modify them
```
tomneedham: indent

tests/Integration/run-test.sh
```diff
@@ -31,6 +31,8 @@ if [ $CODE -eq 0 ]; then
     # so we add an extra check whether tests really succeeded
     if [[ $OUT != *"Tests succeeded"* ]]; then
         echo "a superordinate error occurred – please check the owncloud.log"
+        echo "disaplying the last 50 lines of owncloud log"
+        tail -n 50 ../../../../data/owncloud.log
```
tomneedham: should remove this

lib/Migrations/Version20170927125822.php
patrickjahns: Caching shouldn't be that much of an issue for appliance use case - they will most likely not have many users/groups. But we need to make sure that #125 and this work together when #125 gets merged

tomneedham:
Revive from the dead if we need this later.

PR #137, opened 2017-11-02
Do not set false quota on login

butonic:
When no quota attribute has been configured and no quota default for ldap users, logging in a user will use `false` to set the quota, which becomes ` B`, an invalid quota oc treats as 0b. any existing quota that may have been set by the admin gets overwritten.

This pr only updates the quota of an account if it has been confugired and returns a non `false` value.


PVince81:
Adding @jvillafanez for review as he reworked the default quota logic

PR #138, opened 2017-11-06
Fix working with LDAP replica server

jvillafanez:
Manual port of https://github.com/owncloud/core/pull/27598

Original issue: https://github.com/owncloud/core/issues/27575

Not tested yet with 10.0 (the original patch was partially tested with 9.1 -> https://github.com/owncloud/core/pull/27598#issuecomment-292905373)

### DIFF
--- a/lib/Connection.php
+++ b/lib/Connection.php
@@ -528,7 +528,13 @@ class Connection extends LDAPUtility {
 				) {
 					$this->doConnect($this->configuration->ldapHost,
 						$this->configuration->ldapPort);
-					$bindStatus = $this->bind();
+					if ($this->ldap->isResource($this->ldapConnectionRes)) {
+						$bindStatus = @$this->ldap->bind($this->ldapConnectionRes,
+								$this->configuration->ldapAgentName,
+								$this->configuration->ldapAgentPassword);
+					} else {
+						$bindStatus = false;
+					}
 					$error = $this->getLDAP()->isResource($this->ldapConnectionRes) ?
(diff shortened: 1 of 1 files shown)


butonic:
ok, works, but the `bind()` method was binding twice. I simplified the connection establishment further to fix that.

sync & login with multiple servers works.

lib/Connection.php
```diff
@@ -520,42 +523,51 @@ private function establishConnection() {
 				}
 			}
 
-			$bindStatus = false;
-			$error = -1;
```
butonic: the whole connect logic can be done without the tracking of bind status er error codes ...

lib/Connection.php
```diff
@@ -580,37 +592,26 @@ private function doConnect($host, $port) {
...
-			return false;
-		}
-		$getConnectionResourceAttempt = true;
+
 		$cr = $this->getConnectionResource();
-		$getConnectionResourceAttempt = false;
 		if(!$this->getLDAP()->isResource($cr)) {
 			return false;
 		}
-		$ldapLogin = @$this->getLDAP()->bind($cr,
```
jvillafanez: Bind function not doing `bind()`? This needs a comment: binding is done via `getConnectionResource()` (if I'm not mistaken)

jvillafanez:
No further complains as long as it works.

Note that we might need to backport the commit to stable9.1 for the sake fo consistency.

tomneedham:
rebased, tested with replica docker server locally, killed the original container and with this PR everything goes back to normal :)

PR #145, opened 2017-11-16
Add codecov.io

DeepDiver1975:


PVince81:
MEEP

> No coverage uploaded for pull request base

DeepDiver1975:
> MEEP

yes - as expected - this is the first commit which has coverage - the base commit has no coverage.

PR #151, opened 2017-11-22
fake set displayname support

butonic:
core will only uppdate the account if the backend supports setting it. this is a workaround. needs fixing in core. but core should handle the sync update anyway.


butonic:
I guess this needs to also meke it into the release branch

butonic:
superseded by https://github.com/owncloud/core/pull/29669, but still needs to go in to make it work with 10.0.4

tomneedham:
already in release branch - will be merged back in

PR #156, opened 2017-11-29
Sync should happen in core

tomneedham:


tomneedham:
Core now takes care of this, after check password the accounts is upserted

tomneedham:
Since we already retrieved the user from LDAP when core does the sync it can use the cached UserEntry

tomneedham:
Note: this would target >10.0.5

tomneedham:
Should this kill setupAccount as well?

butonic:
@tomneedham setupAccount is code in core

tomneedham:
This needs to target the version that https://github.com/owncloud/core/pull/30638/commits is released in (10.0.8)

butonic:
faling stable10 tests until https://github.com/owncloud/core/pull/30638 is merged

tomneedham:
@ownclouders rebase please

tomneedham:
Well, I rebased it

tomneedham:
Just testing this with current stable10 HEAD @ d0ee6a6a403e98a106a61b8f7dfcd06313e0b36c

butonic:
during testing I was worried that we removed the updateGroups call ... turns out the cache for getMemberOf is never read anyway. also, the group backend queries memberof itself and resolves nested groups. Tracking caching of groups and primarygroup in https://github.com/owncloud/user_ldap/issues/226

PR #157, opened 2017-12-11
fake set displayname support

butonic:
release port of https://github.com/owncloud/user_ldap/pull/151


tomneedham:
Makes sense because at the moment in core there is a bug with the user->setDisplayName() which only allows setting of the displayname if the backend supports it. This combined with the sync logic means that the displayname from backends does not get updated unless the backend tells core that it supports changing displaynames on the backend directly.

Core logic will be fixed in 10.0.5 but this is needed to fix ldap displayname sync until then.

PR #158, opened 2017-12-11
use custom uuid attribute if configured

butonic:
fixes syncing when a custom uuid attribute is configured
- [ ] needs port to master after release


tomneedham:
>needs port to master after release

according to @patrickjahns we should merge in the release brnach back into master to 'forward port' the small fixes. you would only also commit to master if its blocking dev

Issue #177, opened 2018-02-05
When using encryption, ldap users login for the first time have invalid private keys.


SergioBertolinSG:
### Steps to reproduce

1. Enable encryption.
2. Set up an ldap server - Do not run the user:sync command.
3. Login with an ldap user.

### Expected behaviour
User has new keys created and can use the server without any change required as any other user.

### Actual behaviour
The user gets a warning telling that his private keys are invalid.

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.1

**Cache**
DB

**ownCloud version:** (see ownCloud admin page)
{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.6.1","versionstring":"10.0.6","edition":"Community","productname":"ownCloud"}

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
No.

**Are you using encryption:** 
Yes.

**Logs**

```
{"reqId":"lNA9RDe0Iiyln5CpBG3A","level":2,"time":"2018-02-05T11:39:13+00:00","remoteAddr":"HIDDEN_IP","user":"--","app":"OC\\User\\Manager","method":"POST","url":"\/index.php\/login","message":"User backend OCA\\User_LDAP\\User_Proxy provided no home for <b8e1702e-8742-1034-8cc0-39af73a96b9a>, using <\/opt\/owncloud\/data\/b8e1702e-8742-1034-8cc0-39af73a96b9a>."}
{"reqId":"lNA9RDe0Iiyln5CpBG3A","level":2,"time":"2018-02-05T11:39:16+00:00","remoteAddr":"HIDDEN_IP","user":"b8e1702e-8742-1034-8cc0-39af73a96b9a","app":"no app in context","method":"POST","url":"\/index.php\/login","message":"Could not decrypt the private key from user \"b8e1702e-8742-1034-8cc0-39af73a96b9a\"\" during login. Assume password change on the user back-end. Error message: Bad Signature"}
```

### Client configuration

**Browser**
N/A

SergioBertolinSG:
Working fine using server 10.0.1.

SergioBertolinSG:
It is not possible for the ldap users to set the private key after the first login:

```
[...]

PR #178, opened 2018-02-06
pass password as it is needed by the encryption

butonic:
fixes https://github.com/owncloud/user_ldap/issues/177 if core does not do the sync

### DIFF
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -332,16 +332,17 @@ class Manager {
 	/**
 	 * TODO sync in core
 	 * @param UserEntry $userEntry
+	 * @param string $password
 	 * @return bool
 	 */
-	public function updateAccount (UserEntry $userEntry) {
+	public function updateAccount (UserEntry $userEntry, $password) {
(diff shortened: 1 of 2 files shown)


SergioBertolinSG:
Trying to apply this after applying #156 fails.

```
patching file lib/User/Manager.php
Hunk #1 FAILED at 332.
1 out of 1 hunk FAILED -- saving rejects to file lib/User/Manager.php.rej
patching file lib/User_LDAP.php
Hunk #1 FAILED at 149.
1 out of 1 hunk FAILED -- saving rejects to file lib/User_LDAP.php.rej
```

PVince81:
@SergioBertolinSG in any case this tells me that we'll need a new release of this app to fix the encryption problem. Not a blocker for 10.0.7 then.

tomneedham:
@SergioBertolinSG you do not need to apply 156 first. 156 is giving compatibility with the sync refactoring we did for the next core release. This fix is for ldap to fix the problems in a ldap release .

PR #180, opened 2018-02-08
run integration tests in drone

individual-it:
move the integration tests to drone, including a small fix to the tests themself


patrickjahns:
I've just checked the runtimes on drone - right now it does not make too much sense to have all the individual test run in separate processes.

Most of the tests run with < 10s - so the setup step is what takes most of the time for theses tests to actually start.

Maybe we could actually put a proper TestFramework around them? PHPunit should work fine

PR #185, opened 2018-02-21
allow json output format

individual-it:
for other commands we have the `--output` switch to be able to output the data as json
This switch would be also handy for user_ldap app, to be able to parse that output correctly. This is useful for automated tests

### DIFF
--- a/lib/Command/ShowConfig.php
+++ b/lib/Command/ShowConfig.php
@@ -24,16 +24,16 @@
 
 namespace OCA\User_LDAP\Command;
 
-use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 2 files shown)


lib/Command/ShowConfig.php
```diff
@@ -85,26 +87,37 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+						$value = implode(';', $value);
+					}
+					$rows[] = [$key, $value];
 				}
-				$rows[] = [$key, $value];
+			
+				$table->setRows($rows);
+				$table->render();
+			}
+			else {
```
phil-davis: formatting here? ``else`` is on a new line

PR #188, opened 2018-02-23
reallow avatars to be changed if not provided by ldap

butonic:
Currently, a user cannot change his avatar. Even if ldap does not provide an avatar.
The handleRequest stuff cannot properly be used for methods that return a boolean.

### DIFF
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -236,7 +236,7 @@ class User_Proxy extends Proxy implements
 	 * @throws DoesNotExistOnLDAPException
 	 */
 	public function canChangeAvatar($uid) {
-		foreach($this->backends as $backend) {
+		foreach ($this->backends as $backend) {
 			if ($backend->userExists($uid)) {
 				return $backend->canChangeAvatar($uid);
 			}
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -27,6 +27,8 @@
 
 namespace OCA\User_LDAP;
 
+use OC\ServerNotAvailableException;
+use OCA\User_LDAP\Exceptions\DoesNotExistOnLDAPException;
 use OCP\IConfig;
 use OCP\IUserBackend;
 use OCP\User\IProvidesEMailBackend;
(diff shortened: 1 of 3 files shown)


PVince81:
discuss if still relevant.
if yes, provide enough context for taking over

butonic:
This was found while sitting next to a customer. The problem is that they want to allow users to upload an avatar because they don't store them in ldap. But the logic won't allow that because it confuses true with false ...

butonic:
I'll rebase after going through assigned issues if time permits this week

butonic:
rebased. however this will need documentation. maybe even a config option to prevent ldap users from changing their avatar. I know a lot of ldap servers dont provide avatars ... yet they are useful to recognize users in the sharing dialog ... we should encourage avatars and allow ldap users to upload their own avatar if the ldap server does not provide one.

cc @pmaier1

PVince81:
a config switch per backend or general ?

butonic:
per backend ... in the vue branch per user mapping. there it should be an expert config option.

tomneedham:
IMO config option can come later..?

tomneedham:
Fixing code style failure...

tomneedham:
Should be green now.

PVince81:
add unit tests for the proxy logic at least ?

if needed could be done by someone else, let me know @butonic

PVince81:
@jvillafanez please take over add tests

jvillafanez:
It isn't possible to add unittests in the `User_Proxy` class. The constructor creates a new `User_LDAP` object, which we won't be able to mock, and it calls the `$this->getAccess` method creating quite a bunch of object that we won't be able to mock neither.

For the `checkAccess` tests, I think the function should be removed. The only place where it is checked is in https://github.com/owncloud/user_ldap/blob/7885bd9a1331a11a15fb3305bc01a6589a8bcad9/lib/User/Manager.php#L224 . Such function is called only inside the class (not sure why it's public then) and the call is made after other calls to the unprotected `$this->access`. This means that the `BadMethodCallException` won't be ever called since it's expected to crash at some point earlier by other reasons.

lib/User_Proxy.php
```diff
@@ -224,11 +226,22 @@ public function getDisplayName($uid) {
...
 	 * @param string $uid the ownCloud user name
 	 * @return bool either the user can or cannot
+	 * @throws \OutOfBoundsException
+	 * @throws \InvalidArgumentException
+	 * @throws \BadMethodCallException
+	 * @throws ServerNotAvailableException
+	 * @throws DoesNotExistOnLDAPException
 	 */
 	public function canChangeAvatar($uid) {
-		return $this->handleRequest($uid, 'canChangeAvatar', [$uid]);
```
jvillafanez: I'll need confirmation if this is the way we want to handle this.

With multiple servers, the `handleRequest` includes logic to access directly to the server where the user is. If we have servers 1, 2 and 3, and we know the user is in server 3, the `handleRequest` will access directly to the server 3 without accessing to 1 and 2.
In that scenario, this code will still hit server 1 and 2 always.

jvillafanez: I've tried to restore this by returning null as well as a random number if the user doesn't exists, but it still doesn't work. It seems there are more things to take into account

jvillafanez:
@PVince81 I think this is good to go. As said, I don't think we can add unittests here, and the code improvements might be quite big (might require a big refactor since there are several things that doesn't seem to work properly)

PR #189, opened 2018-02-26
use php7.1 in travis tests

individual-it:
after https://github.com/owncloud/core/pull/28043 php 7.1 has to be used for the tests

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,6 +44,9 @@ pipeline:
     db_host: mysql:3306
     db_username: owncloud
     db_password: owncloud
+    when:
+      matrix:
+        NEED_SERVER: true
 
   owncloud-log:
     image: owncloud/ubuntu:16.04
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
Please add to the test matrix a configuration for 5.6 and 7.0 which executes with CORE_BRANCH=stable10 - this app is to be run on 10.0 as well .... THX

individual-it:
true,
and it does not run on master anyway yet 
```
  App "LDAP Integration" cannot be installed because the following dependenci  
  es are not fulfilled: ownCloud 10.0.99 or lower is required.
```
but that is for a separate PR

individual-it:
1. do not run integration tests on travis (try run fine in drone since a while)
2. run unit tests on travis with stable10 core branch and various PHP versions
3. do not run unit tests with core master for now as the app is not updated to work with master yet
4. run integration tests on drone with stable10 & php 5.6
5.  do not run integration tests on drone with master  for now as the app is not updated to work with master yet
6. run acceptance tests on drone with stable10 and PHP 5.6
7. run acceptance tests on drone with stable10 and PHP 7.0
8. do not run acceptance tests with core master for now as the app is not updated to work with master yet

patrickjahns:
The question for me is, if it rather makes sense to shift running tests completely in drone?

individual-it:
Yes, can do that, only unit tests left

individual-it:
new PR in #192

PR #190, opened 2018-02-26
[WIP] Compatibility with central group and membership table

mrow4a:
- Add is sync maintained flag
 
@butonic @DeepDiver1975 @tomneedham @PVince81

### DIFF
--- a/lib/Group_LDAP.php
+++ b/lib/Group_LDAP.php
@@ -910,4 +910,14 @@ class Group_LDAP implements \OCP\GroupInterface {
 	public function isVisibleForScope($scope) {
 		return true;
 	}
+
+	/**
+	 * Returns whether the groups and memberships are to be maintained by core
+	 * and sync mechanism to be used.
+	 *
+	 * @since 10.0.0
+	 */
+	public function isSyncMaintained() {
+		return true;
+	}
 }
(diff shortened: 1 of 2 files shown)


lib/Group_LDAP.php
```diff
@@ -910,4 +910,14 @@ public function implementsActions($actions) {
 	public function isVisibleForScope($scope) {
 		return true;
 	}
+
+	/**
+	 * Returns whether the groups and memberships are to be maintained by core
+	 * and sync mechanism to be used.
+	 *
+	 * @since 10.0.0
```
tomneedham: ? should this be a reference to cores version, and hence 10.1? or ldap version?

lib/Group_Proxy.php
```diff
@@ -207,4 +207,14 @@ public function implementsActions($actions) {
 	public function isVisibleForScope($scope) {
 		return true;
 	}
+
+	/**
+	 * Returns whether the groups and memberships are to be maintained by core
+	 * and sync mechanism to be used.
+	 *
+	 * @since 10.0.0
```
tomneedham: as above

lib/Group_LDAP.php
mrow4a: the same  as for isVisibleForScope

butonic: the since annotation is used in core to indicate when an API has been introduced. see `\OCP\GroupInterface`.  So this occurence, https://github.com/owncloud/core/pull/29107/files#diff-2764c6ee76270bcf82f6a5fd186b0ae1R133 and probably other `@since` annotations in that PR should be updated to 10.1. If we backport to 10.0.?

mrow4a:
The flag not needed - explanation here https://github.com/owncloud/core/pull/29107

PR #194, opened 2018-02-27
Stop being distracted by annoyances in access.php

butonic:
These tiny things keep annoying me ... get the issues  down a little. one class after another.

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -59,7 +59,7 @@ class Access implements IUserTools {
 	/**
 	 * @var string[] $cookies an array of returned Paged Result cookies
 	 */
-	protected $cookies = array();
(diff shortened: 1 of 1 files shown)


lib/Access.php
```diff
@@ -1296,15 +1338,14 @@ public function combineFilterWithOr($filters) {
 	 * @return string the combined filter
 	 */
 	private function combineFilter($filters, $operator) {
-		$combinedFilter = '('.$operator;
+		$combinedFilter = 	"($operator";
```
tomneedham: random tab?

individual-it:
Note: acceptance tests in drone fail because of https://github.com/owncloud-ci/core/issues/3

PR #199, opened 2018-03-02
[WIP] convert ajax/* to Controllers

butonic:
kill-all-the-legacy

can be further improved by touching js. separate PR


lib/Controller/MappingController.php
```diff
@@ -0,0 +1,92 @@
...
+
+	/**
+	 * test the given ldap config
+	 *
+	 * @param string $ldap_serverconfig_chooser config id // FIXME remove, is unneeded
+	 * @param string $ldap_clear_mapping subject, 'group' or 'user'
+	 * @return DataResponse
+	 */
+	public function clear($ldap_serverconfig_chooser = null, $ldap_clear_mapping) {
+		$subject = $ldap_clear_mapping;
```
jvillafanez: needed?

lib/Controller/WizardController.php
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param string $ldap_serverconfig_chooser config id
+	 * @param string $action the spell to cast
+	 * @param string $cfgkey optional
+	 * @param string $cfgval optional
+	 * @param string $ldap_test_loginname optional
+	 * @return DataResponse
+	 */
+	public function cast($ldap_serverconfig_chooser, $action, $cfgkey = null, $cfgval = null, $ldap_test_loginname = null) {
+		$prefix = $ldap_serverconfig_chooser;
```
jvillafanez: needed?

jvillafanez:
I guess most of the code has been just moved to a controller without any additional change. If so, I think it's fine for a first iteration.

lib/Controller/WizardController.php
butonic: for now. I don't want to touch js with this PR. will remove in a subsequent pr. hopefully ... depends on how it affects js

lib/Controller/MappingController.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * test the given ldap config
+	 *
+	 * @param string $ldap_serverconfig_chooser config id // FIXME remove, is unneeded
+	 * @param string $ldap_clear_mapping subject, 'group' or 'user'
+	 * @return DataResponse
+	 */
+	public function clear($ldap_serverconfig_chooser = null, $ldap_clear_mapping) {
```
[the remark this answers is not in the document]
butonic: testet by clicking in the web ui to copy and create new configs. does work. but I agree this is ugly.

lib/Controller/ConfigurationController.php
```diff
@@ -0,0 +1,205 @@
...
+	 * create a new ldap config
+	 *
+	 * @param string $copyConfig copy the config
+	 * @return DataResponse
+	 */
+	public function create($copyConfig = null) {
+		$prefixes = $this->helper->getServerConfigurationPrefixes();
+		sort($prefixes);
+		$maxPrefix = array_pop($prefixes);
+		$count = (int)ltrim($maxPrefix, 's');
```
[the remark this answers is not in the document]
butonic: yep, see https://github.com/owncloud/user_ldap/pull/199/files#diff-4f4649e82f381345e573748cf8c24c55R89

jvillafanez:
I see several `new Configuration(..)` and `new Connection(..)`. We should consider to inject a factory in those classes. This should make it easier to test those classes.

butonic:
@jvillafanez I agree ... but I am already struggling to keep the changes low ... expect more PRs that will clean this up. The Wizard gets passed an ldap wrapper and an Access instance ... but it actually expects the ldap connection to be the same. If we clean that up DI becomes a lot cleaner ... AND we can get rid of that LDAPUtility class crap.

This was my last commit ... to this PR. Waiting for CI, test wizard manually, then merge ASAP.

butonic:
@individual-it I renamed `tests/Integration` to `tests/integration`. IMO the tests/* folders should not be part of the class names. The unit test classes eg should follow the same namespace as the classes under test. So the class under test does not need an additional use statement.

I hope that is fine for you. Objections?

individual-it:
:+1:  for the integration folder renaming. This are real integration tests and that's great!

I don't really care about the namespace but I feel it should be uniform across oC. Looking in core and in some Apps it seems like most places have `\Tests` at the end of the namespace
Also unit tests mostly do not have a separate sub-folder under tests. Actually I like the idea of having 3 subfolders under tests (unit, integration, acceptance) but it would help future developers and those who switch between apps if everything is uniform

butonic:
> but it would help future developers and those who switch between apps if everything is uniform
@individual-it only if the defaults are sane ... I think namespaces ending in `\Tests` are wrong because I am used to the Java world, where the test classes are placed in the same namespace because if no visibility is given for functions (`public`, `private`, `protected`) the visibility is actually `package-private` (a concept that [does not exist in php](https://bugs.php.net/bug.php?id=55331)). Also see https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

In any case it requires writing less code and I prefer to do things the right way™️.

patrickjahns:
@butonic 
👎 

please set according milestones ....

> In any case it requires writing less code and I prefer to do things the right way™️.

In the PHP world - very often Tests have a namespace prefix of `\Test\.....` - I am used to that - namespace (packagename) as separation to see with what class are you currently working.

But it doesn't matter - "right way™️" is project specific - and please if there is a need to clarify how things should be done - then we add it to the documentation/handbook

Lets please unify code styling in all apps and not go different paths

tomneedham:
if milestone == null, milestone = development ?

PR #200, opened 2018-03-02
check estimate to fix count when it reports the count for all users, …

butonic:
…not the users on the current page

Happens if the ldap server has no page limit configured. Then the count will be (total count)*(no of pages).

This PR also saves a few requests to iterate over all pages if the server provides an estimate.

cc @esakol would be great to hear if this still works as expected or RHDS (it should)
cc @kiu-ball please check if this solves the count, even when ldap page size in oc is 500

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1056,7 +1056,7 @@ class Access implements IUserTools {
 			foreach ($sr as $key => $res) {
 				if ($this->getLDAP()->controlPagedResultResponse($cr, $res, $cookie, $estimated)) {
 					$range = $offset . "-" . ($offset + $limit);
-					$estimates[] = $estimated;
+					$estimates[$key] = $estimated;
 					\OC::$server->getLogger()->debug(
 						'Page response cookie '.$this->cookie2str($cookie)." at $range, estimated<$estimated>",
 						['app' => 'user_ldap']);
@@ -1144,7 +1144,7 @@ class Access implements IUserTools {
 
 			foreach ($counts as $i => $count) {
 				$estimate = $estimates[$i];
-				if ($count === $estimate) {
+				if ($estimate > 0) {
 					// estimate reported for complete result, use it
 					$counter += $estimate;
 					// stop counting entries on subsequent pages for the base with an estimate
@@ -1174,9 +1174,9 @@ class Access implements IUserTools {
 		$cr = $this->connection->getConnectionResource();
 		$counts = [];
 
-		foreach ($searchResults as $res) {
+		foreach ($searchResults as $key => $res) {
 			$count = (int)$this->getLDAP()->countEntries($cr, $res);
-			$counts[] = $count;
+			$counts[$key] = $count;
 		}
 
 		return $counts;
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -2013,13 +2013,7 @@ class Access implements IUserTools {
 						['app' => 'user_ldap']);
 				}
 			}
-			/* ++ Fixing RHDS searches with pages with zero results ++
-			 * We coudn't get paged searches working with our RHDS for login ($limit = 0),
-			 * due to pages with zero results.
-			 * So we added "&& !empty($this->lastCookie)" to this test to ignore pagination
-			 * if we don't have a previous paged search.
-			 */
-		} elseif ($this->connection->hasPagedResultSupport && $limit === 0 && !empty($this->lastCookie)) {
+		} elseif ($this->connection->hasPagedResultSupport && $limit === 0) {
 			// a search without limit was requested. However, if we do use
 			// Paged Search once, we always must do it. This requires us to
 			// initialize it with the configured page size.
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1159,7 +1159,7 @@ class Access implements IUserTools {
 			/* ++ Fixing RHDS searches with pages with zero results ++
 			 * Continue now depends on $hasMorePages value
 			 */
-			$continue = $pagedSearchOK && $hasMorePages && count($base) > 0;
+			$continue = $pagedSearchOK && $hasMorePages && \count($base) > 0;
 		} while ($continue && ($limit === null || $limit <= 0 || $limit > $counter));
 
 		return $counter;
@@ -1256,6 +1256,10 @@ class Access implements IUserTools {
 			|| (!$this->getPagedSearchResultState() && $pagedSearchOK)
 		) {
 			$findings = \array_slice($findings, (int)$offset, $limit);
+		}
+
+		if (!$continue) {
+			$this->abandonPagedSearch();
 		}
 
 		if ($attr !== null) {
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -995,7 +995,7 @@ class Access implements IUserTools {
 	 * second | false if not successful
 	 * @throws \OC\ServerNotAvailableException
 	 */
-	private function executeSearch($filter, $base, &$attr = null, $limit = null, $offset = null) {
+	private function executeSearch($filter, array $base, &$attr = null, $limit = null, $offset = null) {
 		if ($attr !== null && !\is_array($attr)) {
 			$attr = [\mb_strtolower($attr, 'UTF-8')];
 		}
@@ -1043,17 +1043,20 @@ class Access implements IUserTools {
 	 * @param bool $pagedSearchOK whether a paged search has been executed
 	 * @param bool $skipHandling required for paged search when cookies to
 	 * prior results need to be gained
-	 * @return bool cookie validity, true if we have more pages, false otherwise.
+	 * @return array 0 => bool cookie validity, true if we have more pages, false otherwise.
+	 *               1 => estimate for every result. if 0 server just might not support estimates
 	 * @throws \OC\ServerNotAvailableException
 	 */
 	private function processPagedSearchStatus($sr, $filter, $base, $iFoundItems, $limit, $offset, $pagedSearchOK, $skipHandling) {
 		$cookie = null;
-		$estimated = '';
+		$estimated = 0;
+		$estimates = [];
 		if ($pagedSearchOK) {
 			$cr = $this->connection->getConnectionResource();
 			foreach ($sr as $key => $res) {
 				if ($this->getLDAP()->controlPagedResultResponse($cr, $res, $cookie, $estimated)) {
 					$range = $offset . "-" . ($offset + $limit);
+					$estimates[] = $estimated;
 					\OC::$server->getLogger()->debug(
 						'Page response cookie '.$this->cookie2str($cookie)." at $range, estimated<$estimated>",
 						['app' => 'user_ldap']);
@@ -1063,7 +1066,7 @@ class Access implements IUserTools {
 
 			//browsing through prior pages to get the cookie for the new one
 			if ($skipHandling) {
-				return false;
+				return [false, $estimates];
 			}
 			// if count is bigger, then the server does not support
 			// paged search. Instead, he did a normal search. We set a
@@ -1083,7 +1086,9 @@ class Access implements IUserTools {
 		 * cookie is null, with openldap cookie is an empty string and
 		 * to 386ds '0' is a valid cookie. Even if $iFoundItems == 0
 		 */
-		return !empty($cookie) || $cookie === '0';
+		$validCookie = !empty($cookie) || $cookie === '0';
+
+		return [$validCookie, $estimates];
 	}
 
 	/**
@@ -1101,7 +1106,7 @@ class Access implements IUserTools {
 	 *
 	 * @throws \OC\ServerNotAvailableException
 	 */
-	private function count($filter, $base, $attr = null, $limit = null, $offset = null, $skipHandling = false) {
+	private function count($filter, array $base, $attr = null, $limit = null, $offset = null, $skipHandling = false) {
 		\OC::$server->getLogger()->debug(
 			'Count filter:  '.\print_r($filter, true),
 			['app' => 'user_ldap']);
@@ -1130,16 +1135,31 @@ class Access implements IUserTools {
 			 * countEntriesInSearchResults() method signature changed
 			 * by removing $limit and &$hasHitLimit parameters
 			 */
-			$count = $this->countEntriesInSearchResults($sr);
-			$counter += $count;
+			$counts = $this->countEntriesInSearchResults($sr);
 
-			$hasMorePages = $this->processPagedSearchStatus($sr, $filter, $base, $count, $limitPerPage,
-										$offset, $pagedSearchOK, $skipHandling);
+			list($hasMorePages, $estimates) = $this->processPagedSearchStatus(
+				$sr, $filter, $base, $count, $limitPerPage,
+				$offset, $pagedSearchOK, $skipHandling
+			);
+
+			foreach ($counts as $i => $count) {
+				$estimate = $estimates[$i];
+				if ($count === $estimate) {
+					// estimate reported for complete result, use it
+					$counter += $estimate;
+					// stop counting entries on subsequent pages for the base with an estimate
+					// TODO currently all queries search the same ldap server, in theory we could end all here. Not much harm done though
+					unset($base[$i]);
+				} else {
+					$counter += $count;
+				}
+			}
+
 			$offset += $limitPerPage;
 			/* ++ Fixing RHDS searches with pages with zero results ++
 			 * Continue now depends on $hasMorePages value
 			 */
-			$continue = $pagedSearchOK && $hasMorePages;
+			$continue = $pagedSearchOK && $hasMorePages && count($base) > 0;
 		} while ($continue && ($limit === null || $limit <= 0 || $limit > $counter));
 
 		return $counter;
@@ -1147,19 +1167,19 @@ class Access implements IUserTools {
 
 	/**
 	 * @param array $searchResults
-	 * @return int
+	 * @return int[]
 	 * @throws \OC\ServerNotAvailableException
 	 */
 	private function countEntriesInSearchResults($searchResults) {
 		$cr = $this->connection->getConnectionResource();
-		$counter = 0;
+		$counts = [];
 
 		foreach ($searchResults as $res) {
 			$count = (int)$this->getLDAP()->countEntries($cr, $res);
-			$counter += $count;
+			$counts[] = $count;
 		}
 
-		return $counter;
+		return $counts;
 	}
 
 	/**
@@ -1213,7 +1233,7 @@ class Access implements IUserTools {
 				$findings = \array_merge($findings, $this->getLDAP()->getEntries($cr, $res));
 			}
 
-			$continue = $this->processPagedSearchStatus($sr, $filter, $base, $findings['count'],
+			list($continue, ) = $this->processPagedSearchStatus($sr, $filter, $base, $findings['count'],
 								$limit, $offset, $pagedSearchOK,
 										$skipHandling);
 			$offset += $limit;
--- a/lib/ILDAPWrapper.php
+++ b/lib/ILDAPWrapper.php
@@ -166,7 +166,7 @@ interface ILDAPWrapper {
 
 	/**
 	 * Search LDAP tree
-	 * @param resource|resource[] $link LDAP link resource
+	 * @param resource $link LDAP link resource
 	 * @param string $baseDN The DN of the entry to read from
 	 * @param string $filter An LDAP filter
 	 * @param array $attr array of the attributes to read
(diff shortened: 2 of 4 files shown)


butonic:
setting to sev medium because we dont want admins to believe oc does not know how to count users ... whet else might be wrong.

PVince81:
@butonic @tomneedham keep or close?

butonic:
keep, saves iterating over all users during sync only to calculate the max if server gives an estimate.

PVince81:
ok, assuming takeover is straightforward

lib/Access.php
```diff
@@ -1162,36 +1167,51 @@ private function count($filter, $base, $attr = null, $limit = null, $offset = nu
...
+			$counts = $this->countEntriesInSearchResults($sr);
+
+			list($hasMorePages, $estimates) = $this->processPagedSearchStatus(
+				$sr, $filter, $base, $count, $limitPerPage,
+				$offset, $pagedSearchOK, $skipHandling
+			);
+
+			foreach ($counts as $i => $count) {
+				$estimate = $estimates[$i];
+				if ($count === $estimate) {
```
jvillafanez: Is this working as we want? I mean, assuming we have 4 pages, count will have the first 500 entries (if we have a page size of 500) but the estimate is expected to be 1777, for example. The only case where this could be useful is when all the results fits in one page, but in that case we should have already counted all the results in `countEntriesInSearchResults` above.
If the estimate is different than the count, we're still counting all the results.

jvillafanez:
I'm specially worried about the terminology. quoting http://www.faqs.org/rfcs/rfc2696.html (which is linked from http://php.net/manual/en/function.ldap-control-paged-result-response.php)
> Each time the server returns a set of results to the client when processing a search request containing the pagedResultsControl, the server includes the pagedResultsControl control in the searchResultDone message. In the control returned to the client, the size MAY be set to the server's **estimate** of the total number of entries in the entire result set

What I understand is that the server might return an estimate, which might not be accurate for whatever reason. It might return a 1777 for the estimate, but the entry count could be 1800 or 1700 or 1645. I don't think we should rely on the estimate for counting.

If we want to rely on this, I think we should check all the server implementations we officially support (other implementations might not work properly in this regard) and check if the server returns an accurate / right number in all known scenarios.

butonic:
The estimate is only used to print a n/estimate progressbar when syncing. only visual makeup.

jvillafanez:
I'd rather provide 2 different methods then: one for the "actual" count and another one for the estimation, or at least clarify that the `count` method will provide an estimation, not an exact result.

jvillafanez:
Code in this branch works (checked with openldap, without estimate, and with 389ds, withe estimate), but we still need to update the branch and solve the conflicts. My comments still apply.

jvillafanez:
I'm blocking the PR due to a bug: the `user:sync` command doesn't finish. I noticed this in some previous PR but I thought it would be solved by this PR.
Checked with ownCloud 10.2 + this PR branch. I haven't checked with other ownCloud versions yet.

If I remember correctly, https://github.com/owncloud/user_ldap/pull/200#issuecomment-473300022 was checked with either OC 10.1 or 10.0.10. Minimum OC version has been increased to 10.2 for the user_ldap code in master recently.

The bug might be unrelated to this PR, so it need further checks. It seems the LDAP app is returning duplicated information after looping through all the users: for 1000 users, ranges 0-500 and 500-1000 seems fine, but it somehow go through range 1000-1500, 1500-2000, etc. I've just checked the first and last user ids of the range, and 0-500, 1000-1500, 2000-2500, etc have the same ids; the same for ranges 500-1000, 1500-2000, 2500-3000, etc.

jvillafanez:
This should be ready. Maybe @butonic can double-check my changes. I particularly don't like 
https://github.com/owncloud/user_ldap/pull/200/commits/ea32a34e485cc5a8d837a2be10760312a5bc22dd but otherwise there is an infinite loop due to the page cookie not being cleanup properly and the LDAP server returns duplicated results. Not sure where to place that check other than there.

jvillafanez:
> I'm blocking the PR due to a bug

This seems to happen only with OC 10.0.10 but not with OC 10.1

jvillafanez:
I'll need help of @owncloud/qa here. I've tried to reproduce the problem that happens in drone manually, but it works on my end. I'm also having trouble executing the acceptance tests (checked the acceptance-ldap-webUI) because the tests seem to try to insert the same LDAP data for each test, and the test fails because the data already exists.

jvillafanez:
@phil-davis @individual-it  I need confirmation that the failures are unrelated to the code changes, or maybe there is a problem in the tests themselves.
```
Feature: Sharing between local and LDAP users
  As a user
  I want to be able to share files and folders with any user regardless of the backend

[Thu Mar 28 11:34:39 2019] 127.0.0.1:36230 [200]: /ocs/v2.php/apps/testing/api/v1/occ
[Thu Mar 28 11:34:39 2019] 127.0.0.1:36248 [200]: /ocs/v2.php/apps/testing/api/v1/opcache
[Thu Mar 28 11:34:39 2019] 127.0.0.1:36258 [200]: /ocs/v1.php/cloud/capabilities
[Thu Mar 28 11:34:39 2019] 127.0.0.1:36268 [200]: /ocs/v1.php/apps/testing/api/v1/apps
[Thu Mar 28 11:34:40 2019] 127.0.0.1:46436 [200]: /ocs/v1.php/cloud/capabilities
[Thu Mar 28 11:34:40 2019] 127.0.0.1:46446 [200]: /ocs/v1.php/apps/testing/api/v1/apps
[Thu Mar 28 11:34:40 2019] 127.0.0.1:36298 [200]: /ocs/v1.php/cloud/apps?filter=enabled
[Thu Mar 28 11:34:40 2019] 127.0.0.1:36308 [200]: /ocs/v1.php/cloud/apps?filter=disabled
  Background:                                                                                     # /var/www/owncloud/apps/user_ldap/tests/acceptance/features/apiUserLDAPSharing/sharingLocalUserLdapUser.feature:6
[Thu Mar 28 11:34:40 2019] User backend OC\User\Database is returning home: /mnt/data/files/local-user for user: local-user which differs from existing value: 
[Thu Mar 28 11:34:40 2019] 127.0.0.1:36318 [200]: /ocs/v2.php/cloud/users
[Thu Mar 28 11:34:40 2019] 127.0.0.1:36328 [200]: /ocs/v2.php/cloud/users/local-user
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36338 [200]: /ocs/v2.php/cloud/users/local-user
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36348 [200]: /ocs/v1.php/cloud/users/local-user
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36358 [200]: /ocs/v2.php/cloud/users/local-user
    Given user "local-user" has been created with default attributes in the database user backend # FeatureContext::userHasBeenCreatedOnDatabaseBackend()
[Thu Mar 28 11:34:41 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36368 [401]: /ocs/v1.php/cloud/users/user0
[Thu Mar 28 11:34:41 2019] Login failed: 'user1' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36378 [401]: /ocs/v1.php/cloud/users/user1
[Thu Mar 28 11:34:41 2019] Login failed: 'user2' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:41 2019] 127.0.0.1:36388 [401]: /ocs/v1.php/cloud/users/user2
    And these users have been initialized:                                                        # FeatureContext::theseUsersHaveBeenInitialized()
      | username |
      | user0    |
      | user1    |
      | user2    |

  Scenario: Share a folder from an LDAP user to a local user                               # /var/www/owncloud/apps/user_ldap/tests/acceptance/features/apiUserLDAPSharing/sharingLocalUserLdapUser.feature:15
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] No cached ldap result found for LOGINNAME,user0
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] 127.0.0.1:36398 [401]: /ocs/v1.php/apps/files_sharing/api/v1/shares?path=%2FPARENT
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] No cached ldap result found for LOGINNAME,user0
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] 127.0.0.1:36408 [401]: /ocs/v1.php/apps/files_sharing/api/v1/shares
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] No cached ldap result found for LOGINNAME,user0
[Thu Mar 28 11:34:42 2019] Login failed: 'user0' (Remote IP: '127.0.0.1')
[Thu Mar 28 11:34:42 2019] 127.0.0.1:36418 [401]: /ocs/v1.php/apps/files_sharing/api/v1/shares?path=%2FPARENT
    When user "user0" shares folder "/PARENT" with user "local-user" using the sharing API # FeatureContext::userSharesFileWithUserUsingTheSharingApi()
[Thu Mar 28 11:34:42 2019] 127.0.0.1:36428 [404]: /remote.php/webdav/PARENT%20(2)/parent.txt
    Then the content of file "/PARENT (2)/parent.txt" for user "local-user" should be:     # FeatureContext::contentOfFileForUserShouldBePyString()
      """
      ownCloud test text file parent
      
      """
      Failed asserting that two strings are equal.
      --- Expected
      +++ Actual
      @@ @@
      -'ownCloud test text file parent
      +'<?xml version="1.0" encoding="utf-8"?>
      +<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
      +  <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
      +  <s:message>File with name PARENT (2)/parent.txt could not be located</s:message>
      +</d:error>
       '
```

Note that there are some login failures and "/ocs/v1.php/cloud/users/user0" and similar calls are returning a 401 code. However, after stopping the tests halfway through, I can login normally with those account via web interface, and I can see all the files.
As said in the comment above, reproducing the test case manually works with any problem.

individual-it:
I will have a look.
There are also integration tests failures https://drone.owncloud.com/owncloud/user_ldap/1313/1131 does that give you any clue?

individual-it:
added to next sprint, for me too look at

jvillafanez:
The problem with the integration tests seems to be cause by my last commit (the one fixing the infinite loop), however, removing that commit causes the infinite loop with a 389ds server when the users are synced.
Infinite loop also seems to happen, at least, with 389ds and master code of the ldap app.

individual-it:
@ownclouders rebase

ownclouders:
Hey! I'm [GitMate.io](https://gitmate.io)! This pull request is being rebased automatically. Please **DO NOT** push while rebase is in progress or your changes would be lost permanently :warning:

ownclouders:
Automated rebase with [GitMate.io](https://gitmate.io) was successful! :tada:

individual-it:
one thing why tests are failing are that an LDAP use cannot access `ocs/v1.php/cloud/users/<user-name>`
e.g. with a user created by occ or API I can access `curl -u uu5:uu5 http://172.17.0.1/ocs/v1.php/cloud/users/uu5`
and get:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data>
  <quota>
   <free>134724288512</free>
   <used>354</used>
   <total>134724288866</total>
   <relative>0</relative>
   <definition>default</definition>
  </quota>
  <email/>
  <displayname>uu5</displayname>
  <home>/home/artur/www/owncloud-core/data/uu5</home>
  <two_factor_auth_enabled>false</two_factor_auth_enabled>
 </data>
</ocs>
```

with an LDAP user:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>997</statuscode>
  <message>Unauthorised</message>
 </meta>
 <data/>
</ocs>
```

in the tests we use that endpoint to make sure the user was initialized (skeleton files copied)

jvillafanez:
I guess the error comes from https://github.com/owncloud/core/blob/master/apps/provisioning_api/lib/Users.php#L114 (there is another error thrown above but it's thrown if the user isn't logged in, which shouldn't be the case).
Looks like that endpoint is only available for admins.

individual-it:
@jvillafanez you are saying it also should not work with non-LDAP users? because it does

jvillafanez:
ok, wrong troubleshooting (I was checking a slightly different endpoint).
The problem should be around https://github.com/owncloud/core/blob/stable10/apps/provisioning_api/lib/Users.php#L208

I guess the actual call should be something like:
```
curl -u loginname:password http://server.com/ocs/v1.php/cloud/users/uid
```
with the default ldap configuration, I expect something like
```
curl -u user0:123456 http://server.com/ocs/v1.php/cloud/users/a1b2-c678-9988-abcac123
```

individual-it:
sorry for the late reply

my LDAP config is:
```
+-------------------------------+---------------------------------------------------------------------------------------+
| Configuration                 | LDAPTestId                                                                            |
+-------------------------------+---------------------------------------------------------------------------------------+
| hasMemberOfFilterSupport      | 0                                                                                     |
| hasPagedResultSupport         |                                                                                       |
| homeFolderNamingRule          |                                                                                       |
| lastJpegPhotoLookup           | 0                                                                                     |
| ldapAgentName                 | cn=admin,dc=owncloud,dc=com                                                           |
| ldapAgentPassword             | ***                                                                                   |
| ldapAttributesForGroupSearch  |                                                                                       |
| ldapAttributesForUserSearch   |                                                                                       |
| ldapBackupHost                |                                                                                       |
| ldapBackupPort                |                                                                                       |
| ldapBase                      | dc=owncloud,dc=com                                                                    |
| ldapBaseGroups                | dc=owncloud,dc=com                                                                    |
| ldapBaseUsers                 | dc=owncloud,dc=com                                                                    |
| ldapCacheTTL                  | 600                                                                                   |
| ldapConfigurationActive       | 1                                                                                     |
| ldapDynamicGroupMemberURL     |                                                                                       |
| ldapEmailAttribute            | mail                                                                                  |
| ldapExperiencedAdmin          | 0                                                                                     |
| ldapExpertUUIDGroupAttr       |                                                                                       |
| ldapExpertUUIDUserAttr        | uid                                                                                   |
| ldapExpertUsernameAttr        |                                                                                       |
| ldapGroupDisplayName          | cn                                                                                    |
| ldapGroupFilter               | (&(|(objectclass=posixGroup)))                                                        |
| ldapGroupFilterGroups         |                                                                                       |
| ldapGroupFilterMode           | 0                                                                                     |
| ldapGroupFilterObjectclass    | posixGroup                                                                            |
| ldapGroupMemberAssocAttr      | memberUid                                                                             |
| ldapHost                      | localhost                                                                             |
| ldapIgnoreNamingRules         |                                                                                       |
| ldapLoginFilter               | (&(|(objectclass=inetOrgPerson))(|(uid=%uid)(|(mailPrimaryAddress=%uid)(mail=%uid)))) |
| ldapLoginFilterAttributes     |                                                                                       |
| ldapLoginFilterEmail          | 1                                                                                     |
| ldapLoginFilterMode           | 0                                                                                     |
| ldapLoginFilterUsername       | 1                                                                                     |
| ldapNestedGroups              | 0                                                                                     |
| ldapNetworkTimeout            | 2                                                                                     |
| ldapOverrideMainServer        |                                                                                       |
| ldapPagingSize                | 500                                                                                   |
| ldapPort                      | 389                                                                                   |
| ldapQuotaAttribute            |                                                                                       |
| ldapQuotaDefault              |                                                                                       |
| ldapTLS                       | 0                                                                                     |
| ldapUserDisplayName           | displayName                                                                           |
| ldapUserDisplayName2          |                                                                                       |
| ldapUserFilter                | (|(objectclass=inetOrgPerson))                                                        |
| ldapUserFilterGroups          |                                                                                       |
| ldapUserFilterMode            | 0                                                                                     |
| ldapUserFilterObjectclass     | inetOrgPerson                                                                         |
| ldapUserName                  | samaccountname                                                                        |
| ldapUuidGroupAttribute        | auto                                                                                  |
| ldapUuidUserAttribute         | auto                                                                                  |
| turnOffCertCheck              | 0                                                                                     |
| useMemberOfToDetectMembership | 1                                                                                     |
+-------------------------------+---------------------------------------------------------------------------------------+
```

for me it does not look like the problem is about the uid but about the permissions of the user.
requesting `/ocs/v1.php/cloud/users/uid` as admin work as user not

Should the user be able to query its own information? If yes, then the code in this branch should be fixed, if not then we should change it so that its also not possible for a database user and need to adjust the test-code.

jvillafanez:
> for me it does not look like the problem is about the uid but about the permissions of the user.
requesting /ocs/v1.php/cloud/users/uid as admin work as user not

It works for the admin, and for any DB user because the user id is the same as the login. You log in with "user0" and the user id is "user0". This doesn't happen with LDAP: you log in with "user0" and the user id is "a1b2-c678-9988-abcac123"

The configuration you might want to change is the "ldapExpertUsernameAttr" to use the uid.

jvillafanez:
Confirmed a problem regarding login with the last commit. This login problem doesn't happen with the web UI.

* Without the "fix infinite loop" commit: the ldap user can access normally to any service (webdav, ocs, etc). However there is an infinite loop in the user:sync command with 389ds
* With the "fix infinite loop" commit: the ldap user can access through the web UI but not externally with curl. There is no problem syncing 389ds

jvillafanez:
Rebased.

This should be fixed now. Maybe someone can check my last 2 commits and find a better solution.
So far I've checked that the `user:sync` command and the call to the `/ocs/v1.php/cloud/users/uid` works with openLDAP and 389ds against core's master branch.

individual-it:
acceptance tests passing here, so I'm unassigned myself - feel free to ping me if you need help

PVince81:
@VicDeo can you review ?

lib/ILDAPWrapper.php
```diff
@@ -166,7 +166,7 @@ public function read($link, $baseDN, $filter, $attr);
 
 	/**
 	 * Search LDAP tree
-	 * @param resource $link LDAP link resource
+	 * @param resource|resource[] $link LDAP link resource
```
VicDeo: @jvillafanez is it really capable to handle array of resources? As I see calling it will end up in `invokeLDAPMethod` that will execute `ldap_read()` using `resource[]` as the first argument

lib/LDAP.php
```diff
@@ -183,7 +183,7 @@ public function nextEntry($link, $result) {
 	}
 
 	/**
-	 * @param LDAP $link
+	 * @param LDAP|LDAP[] $link
```
VicDeo: the same as above. In addition PHPDoc for the Interface method should be updated accordingly.

lib/LDAP.php
```diff
@@ -194,7 +194,7 @@ public function read($link, $baseDN, $filter, $attr) {
...
```
VicDeo: the same as above.

lib/ILDAPWrapper.php
jvillafanez: fixed

lib/LDAP.php
jvillafanez: fixed

lib/LDAP.php
jvillafanez: fixed

jvillafanez:
Rebased.

jvillafanez:
If the tests pass I think we can merge this in order to move on with the next ldap stuff that is pending.
I don't think we can rise the code coverage in the Access.php file in a reasonable time. The code is really difficult to deal with, it might take a whole dedicated week to add tests for this PR (assuming it's possible). It doesn't seem worthy.

micbar:
@butonic could you clarify how we should proceed here?
Is it worth to add that effort into writing more tests as @jvillafanez proposed?

jvillafanez:
We should, but the code is such a mess that we'd need to refactor it heavily before attempting to add new tests. This is something we can't do within this PR's scope.

VicDeo:
@jvillafanez as all changed methods are not public unit test(s) should be written for `countUsers`

jvillafanez:
For reference (courtesy of http://www.phpmetrics.org)
Screen Shot 2019-05-16 at 12 49 26

butonic:
have a look at where countUsers is actually used. AFAICT it boils down to showing a progress bar. For that we really don't need tests before merging this. If the numbers are off it might be because they are based on an 'estimate'. When the progress bar finishes it will show the correct number of users. Rebase and merge.

PR #197, opened 2018-02-28
[for reference] Detect LDAP Vendor & Version in wizard

butonic:
I need to wrap my head around the wizard and config classes and web ui. This PR adds detection of the LDAP vendor and LDAP version ... if the server announces it. OpenLDAP and Active Directory don't ... of course ... they can only be detected by checking additional attributes

Try these public servers in a new empty config with port 389:
```
ldap://directory.d-trust.de // Oracle Corporation | Sun-Directory-Server/11.1.1.7.2
ldap://x500.bund.de // Atos IT Solutions and Services GmbH | DirX Directory V8.3 8.9.310 2014:09:08 20:10 64-Bit
ldap://directory.verisign.com // Red Hat, Inc. | Red Hat-Directory/8.1.0 B2009.111.1832
ldap://ldap.duke.edu // 389 Project | 389-Directory/1.2.11.15 B2015.022.1831
ldap://ldap.forumsys.com // objectclass=OpenLDAProotDSE
ldap://ldap-tiks.telesec.de // objectclass=OpenLDAProotDSE
ldap://www.trustcenter.de // CentOS | CentOS-Directory/8.2.8 B2012.041.1227
```

AD is detected as well but I dont have one I can post as an example.

- [ ] use span instead of input element
- [ ] add this to the configreport

cc @tomneedham @felixheidecke for insights how to add a single thing to the wizard.

PR #210, opened 2018-03-14
LDAP Wizard rewrite with Vue JS

felixheidecke:
To make configuring ldap servers much less error prone and more self explainable, this PR does two things:
- introduce a new vue.js based wizard
- rewrite the configuration management

## vue.js wizard

The new ui is intended to make configuring ldap easier and less error prone. It presents the user with a list of known configurations, as well as status information like (active, number of users, number of groups ...).

New configuretions can be created or copied from existing ones. An active configuration cannot be edited.
@felixheidecke _actually we can do that now, because the changes are not saved and discovery uses a complete config object to set up the connection / app_

- [ ] add mockups here? the 'flow' is list configs -> new / edit -> configure server -> add user tree -> add group tree. when adding a tree the minimal config just uses the base dn for the tree and the mapping of attributes. filters can then limit the number of users in the tree, by objectclass, group membership, arbitrary attribute value checks ... raw filters

## Configuration model

That leads to the introductuon of a new configuration model:
- a server has some basic options and contains multiple trees.
- every tree needs a base dn and either contains a user or a group mapping.

## migration and breaking change

On the downside it requires migrating old configs (trivial because they can be expressed in the new configuration)
Unfortunately, the new config is based on json, which is a breaking change: rather than multiple key value settings a single object is stored in the config. Following semver, a major version change is needed.

The old `ldap:set-config` is partially compatible. The parameters do not allow us to automatically determine which tree a change belongs to, so we can only make it backward compatible for migrated configs or configs that only have a single user / group tree.

To configure the ldap server you can use `occ ldap:import-config`
```console
$ echo '{
  "id": "exampleconfig-must-be-unique",
  "active": true,
  "host": "openldap",
  "port": 389,
  "bindDN": "cn=admin,dc=owncloudqa,dc=com",
  "password": "secret",
  "supportsPaging": true,
  "supportsMemberOf": true,
  "userTrees": {
    "ou=people,dc=owncloudqa,dc=com": {
      "baseDN": "ou=people,dc=owncloudqa,dc=com",
      "usernameAttribute": "uid",
      "displayName2Attribute": "mail",
      "additionalSearchAttributes": ["cn", "sn"]
    },
    "ou=admins,dc=owncloudqa,dc=com": {
      "baseDN": "ou=admins,dc=owncloudqa,dc=com",
      "usernameAttribute": "uid",
      "displayName2Attribute": "mail",
      "additionalSearchAttributes": ["cn", "sn"]
    }
  },
  "groupTrees": {
    "ou=groups,dc=owncloudqa,dc=com": {
      "baseDN": "ou=groups,dc=owncloudqa,dc=com",
      "filterObjectclass": [
        "groupOfNames"
      ],
      "memberAttribute": "member",
      "displayNameAttribute": "cn"
    }
  }
}' | occ ldap:import-config
```
This will set up two user trees and a single group tree.

## Development

Requires node.js v10 and up.

Run `npm install && npm run build` to compile a working version.

The most interesting route for the ui is the discovery. It takes a server config object as the POST payload and a query string in the url to find 10 matching users / groups. It dues need a requesttoken and a session from the web interface:

```console
/home/jfd$ curl 'http://localhost:8000/oc/core-st/DQEeNSU=:4Ldd0lSCMZSNMrL/KjlR+Xb/6gH5SDXluINoA7e2Fb0=' -H 'Cookie: oc6wa9m1ilwz=6d2ummn7q1stmnugfr6nb8qvon; oc2n5n33xk23=77fcucua6i65j6i8ueck2ljd39; oc_sessionPassphrase=wbkEF4%2FSomCHKjs1mwGTCAnbh9xIUE8%2BBIq29ECNlk9ATuyxoKsdpcfkE0GBjr7totbgeVqolpZbNQUcMX8cWFMnfgiJUvltnrB2fR%2B0wysmxwqx%2Baamm4nTDmfz5nBY; ocy6dyqevwf7=n1cu6tsckojdis47qdmq1701ch' -H "Content-Type: application/json" -d '{
  "id": "exampleconfig-must-be-unique",
  "host": "127.0.0.1",
  "port": 10389,
  "bindDN": "cn=admin,dc=example,dc=org",
  "password": "secret",
  "supportsPaging": true,
  "supportsMemberOf": true,
  "userTrees": {
    "ou=users,dc=example,dc=org": {
      "baseDN": "ou=users,dc=example,dc=org",
      "usernameAttribute": "cn",
      "displayName2Attribute": "mail"
    }
  },
  "groupTrees": {
    "ou=groups,dc=example,dc=org": {
      "baseDN": "ou=groups,dc=example,dc=org",
      "filterObjectclass": ["group"],
      "memberAttribute": "uniquemember",
      "displayNameAttribute": "cn"
    }
  }
}' | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   914  100   305  100   609    663   1325 --:--:-- --:--:-- --:--:--  1326
{
   "cn=zombie1001,ou=users,dc=example,dc=org" : {
      "memberof" : [
         "cn=horde1000,ou=groups,dc=example,dc=org",
         "cn=horde1002,ou=groups,dc=example,dc=org"
      ],
      "entryuuid" : [
         "943c5f20-cc47-4d43-9c26-af297c5150b2"
      ],
      "displayname" : [
         "Bobbie"
      ],
      "mail" : [
         "bobbie@example.org"
      ],
      "cn" : [
         "zombie1001"
      ],
      "objectclass" : [
         "inetOrgPerson"
      ]
   }
}
```


felixheidecke:
We could also wait for `$(document).ready()` to start vue so there is no need to move JS to the footer. But the latter is still recommended.

patrickjahns:
@felixheidecke 
it would be great to add some tests for javascript / vuejs here - I can assist with getting them run in drone

felixheidecke:
@butonic The object returning from the configurations route is weirdly encapsulated if there is no config present:
```
{
  "": {
    "enabled": "yes",
    "has_memberof_filter_support": "0",
    "home_folder_naming_rule": "",
    [...]
  }
}
```

patrickjahns:
@felixheidecke 
please also `git rebase` - merge-commits from remote are highly discouraged https://github.com/owncloud/user_ldap/pull/210/commits/602ce26cdd21e8beff4dbc6c8a4667d11a659c9a

DeepDiver1975:
please address the reported vulnerabilities -- THX

butonic:
I plan to fix https://github.com/owncloud/user_ldap/issues/310 with this as well

butonic:
- [ ] fix existing unit tests
- [ ] fix commands, most of them need a rewrite with backward compatible output
- [ ] add tests to cover multi base dn and multi server configuration scenarios. these corner cases are important
- [ ] adapt integration tests to use new config model
- [ ] add migration for legacy config
- [ ] do a release for a version 2.0

PVince81:
- [ ] consider UI acceptance tests ? (TODO: estimate)

and should it be testing only whether the UI clicks fine or should it also cover multiple LDAP servers ?

PVince81:
- [ ] test / usability test with actual admins

PVince81:
Quoting @butonic from chat:

> the new ui was intended to make configuring ldap easier and less error prone. that lead to the introductuon of a new configuration management. an server server has some basic options and contains multiple trees. every tree needs a base dn and either contains a user or a group mapping. that makes configuring ldap servers much more explainable, and remuves the weird multiline textareas in the wizard. But it also requires migrating old configs (which is trivial) an is a breaking change as the config is now a json object, rather than multiple key value settings. so a semver major version change is needed.

- [ ] plan major release for this change @PVince81

butonic:
to configure the ldap server you can use
```console
$ echo '{
  "id": "s01",
  "active": true,
  "host": "openldap",
  "port": 389,
  "bindDN": "cn=admin,dc=owncloudqa,dc=com",
  "password": "owncloud123",
  "tls": false,
  "turnOffCertCheck": false,
  "supportsPaging": true,
  "pageSize": 500,
  "supportsMemberOf": false,
  "overrideMainServer": false,
  "backupHost": null,
  "backupPort": null,
  "backupTTL": 30,
  "cacheTTL": 600,
  "userTrees": {
    "ou=people,dc=owncloudqa,dc=com": {
      "baseDN": "ou=people,dc=owncloudqa,dc=com",
      "loginFilter": "(&(objectclass=person)(uid=%uid))",
      "loginFilterMode": 0,
      "loginFilterEmail": false,
      "loginFilterUsername": true,
      "loginFilterAttributes": [],
      "usernameAttribute": "uid",
      "expertUsernameAttr": null,
      "displayName2Attribute": "mail",
      "emailAttribute": "mail",
      "homeFolderNamingRule": null,
      "quotaAttribute": null,
      "quotaDefault": null,
      "type": "user",
      "filterObjectclass": [
        "person"
      ],
      "filterGroups": [],
      "filter": "(objectclass=person)",
      "filterMode": 0,
      "uuidAttribute": "auto",
      "displayNameAttribute": "displayName",
      "additionalSearchAttributes": ["cn", "sn"]
    },
    "ou=admins,dc=owncloudqa,dc=com": {
      "baseDN": "ou=admins,dc=owncloudqa,dc=com",
      "loginFilter": "(&(objectclass=person)(uid=%uid))",
      "loginFilterMode": 0,
      "loginFilterEmail": false,
      "loginFilterUsername": true,
      "loginFilterAttributes": [],
      "usernameAttribute": "uid",
      "expertUsernameAttr": null,
      "displayName2Attribute": "mail",
      "emailAttribute": "mail",
      "homeFolderNamingRule": null,
      "quotaAttribute": null,
      "quotaDefault": null,
      "type": "user",
      "filterObjectclass": [
        "person"
      ],
      "filterGroups": [],
      "filter": "(objectclass=person)",
      "filterMode": 0,
      "uuidAttribute": "auto",
      "displayNameAttribute": "displayName",
      "additionalSearchAttributes": ["cn", "sn"]
    }
  },
  "groupTrees": {
    "ou=groups,dc=owncloudqa,dc=com": {
      "baseDN": "ou=groups,dc=owncloudqa,dc=com",
      "filterObjectclass": [
        "groupOfNames"
      ],
      "nestedGroups": false,
      "dynamicGroupMemberURL": null,
      "memberAttribute": "member",
      "filter": "(objectclass=groupOfNames)",
      "filterMode": 0,
      "uuidAttribute": "auto",
      "displayNameAttribute": "cn",
      "additionalSearchAttributes": []
    }
  }
}' | occ ldap:import-config
```

butonic:
The most important controller is `ConfigurationController::discover($query)` which is a post to `configuration/discover/{query}` with the full configuration as payload.

butonic:
rebase to bring in
- [ ] https://github.com/owncloud/user_ldap/pull/324/

butonic:
- [ ] add server vendor and product version detection code from https://github.com/owncloud/user_ldap/pull/197

butonic:
@felixheidecke @PVince81 I rebased and converted the timeout config option from https://github.com/owncloud/user_ldap/pull/324

I don't want to bring in https://github.com/owncloud/user_ldap/pull/197. While we could use the autodetected server to prefill config options I see that as an optimization after we switched to the new UI.

Other than that, the list in https://github.com/owncloud/user_ldap/pull/210#issuecomment-443740166 still applies

PVince81:
I wonder if we should already start branching from the old LDAP so we could merge the current work to master and continue in separate tickets

PVince81:
or maybe we first wait for completion for the remaining PRs...

PVince81:
I think some of the tests (non-unit) should be written by @owncloud/qa while @VicDeo focusses on unit tests and the remaining tasks.

@individual-it @phil-davis is the current LDAP coverage / test code already far enough that we could add tests for the new wizard ?

individual-it:
> @individual-it @phil-davis is the current LDAP coverage / test code already far enough that we could add tests for the new wizard ?

there are no UI LDAP tests at all, the api tests I think have a reasonable coverage

VicDeo:
I should say that backend is still need fixing here and there.
According to my tests it can't connect via `ldap:// ` protocol, and dies with 
`"message":"ldap_start_tls(): Unable to start TLS: Can't contact LDAP server at \/home\/deo\/public_html\/oc-tmp\/apps\/user_ldap\/lib\/LDAP.php#281"}`
in logs

VicDeo:
ok. got it... When UI responds with string 'false' it is saved to DB as `true`:
```
>php -r "var_dump((bool) 'false');"
Command line code:1:
bool(true)
```

In addition old wizard tried to connect with and without TLS and saved this option according to the results. How does this stack with a new UI?

VicDeo:
Error message `Lost connection to LDAP server` while connecting to invalid host/port should say `No connection` instead

VicDeo:
style fixed +2 successful jobs. Now making the phan inspection happy:
```
lib/Command/ShowConfig.php:93 PhanUndeclaredClassMethod Call to method __construct from undeclared class \OCA\User_LDAP\Command\Configuration
lib/Command/ShowConfig.php:94 PhanUndeclaredClassMethod Call to method getConfiguration from undeclared class \OCA\User_LDAP\Command\Configuration
lib/Group_LDAP.php:708 PhanParamTooFew Call with 1 arg(s) to \OCA\User_LDAP\Connection\FilterBuilder::getFilterPartForUserSearch() which requires 2 arg(s) defined at lib/Connection/FilterBuilder.php:171
lib/Group_LDAP.php:807 PhanParamTooFew Call with 1 arg(s) to \OCA\User_LDAP\Connection\FilterBuilder::getFilterPartForUserSearch() which requires 2 arg(s) defined at lib/Connection/FilterBuilder.php:171
```

VicDeo:
### CLI
####  occ ldap:show-config 
- [ ]  what is the correct behavior when config id is unspecified?
current behavior:
```

In ServerMapper.php line 89:
                          
  config- does not exist  
                          

ldap:show-config [--output [OUTPUT]] [--show-password] [--] [<configID>]
```

- [ ] prints nothing with default output format (works with --output=json or --output=json_pretty)
- [ ] always shows clear text password even when there is no `--show-password` argument provided
- [ ] does not support old (non-json) configs.


####  occ ldap:delete-config 
- [ ] does not support old (non-json) configs.


#### occ ldap:search 
- [ ] does not work (probably due to missing displayName attribute mapping)

#### occ ldap:set-config
- [ ] does not validate input keys/values
- [ ] should list all available keys
e.g. 
`php occ ldap:set-config someconfig ldap_port hello`

#### occ ldap:test-config
-  [ ] says everything is valid and ok on
`php occ ldap:set-config someconfig ldap_port hello`
`php occ ldap:test-config someconfig`

VicDeo:
Some notes

#### Autoprobe isTls
- `isTls` could be probed when config is about to be enabled.

#### Migration to json-based configs
- an explicit migration step. the migration should not magically happen in the background or during the upgrade
the legacy wrapper should bu used as long as a legacy config is there but once the config is written / changed it will be stored as json
i would add to the release notes that admins should run an explicit migration command to have their config migrated (if possible)

#### multiple user/group mappings within one config
problem of not knowing which user tree might contain the users for a group:
- https://github.com/owncloud/user_ldap/pull/210/files#diff-7d96cbe30d75b7cc761787a95c100f75R386
- https://github.com/owncloud/user_ldap/pull/210/files#diff-7d96cbe30d75b7cc761787a95c100f75R701
- https://github.com/owncloud/user_ldap/pull/210/files#diff-7d96cbe30d75b7cc761787a95c100f75R802
I think we may need to find users from all user trees. Or we may need to specify which user trees belong to which group.
however, I think by putting a user tree and a group tree in the same config you kind of implicitly say: hey users from these user trees can show up in any of these group trees

DeepDiver1975:
dead -> close

PR #214, opened 2018-03-19
No avatar DI

butonic:
fixes a listing group shares when ldap and encryption are used


PVince81:
Approach sounds good. @jvillafanez please review in detail

jvillafanez:
I won't complain if this is needed for other ownCloud versions, but the solution itself is quite bad.

butonic:
Indeed. Basically a quickfix that we can release without having to QA and release a new version of core. The proper fix is to move avatar sync to core. tracked in https://github.com/owncloud/core/issues/30840

cdamken:
@butonic Can I add the QA Label in both issues?

tomneedham:
Core fix implemented instead - this isn't required https://github.com/owncloud/core/pull/30810

PR #216, opened 2018-03-21
Refactor acceptance tests for core changes

phil-davis:
and other goodies to make acceptance tests work.


phil-davis:
This uses the nightly core build. So the CI needs to be run again once the latest nightly build has happened...

PR #163, opened 2017-12-20
Release/0.10.0

DeepDiver1975:
merge once release ...

PR #218, opened 2018-03-27
Displayname nor email won't be editable from the profile page

jvillafanez:
Fix https://github.com/owncloud/enterprise/issues/2459 by making the profile read-only for the displayname and email.

Consider this PR as **untested** unless told otherwise.

### DIFF
--- a/lib/User_LDAP.php
+++ b/lib/User_LDAP.php
@@ -271,7 +271,6 @@ class User_LDAP implements IUserBackend, UserInterface {
 		return (bool)((Backend::CHECK_PASSWORD
 			| Backend::GET_HOME
 			| Backend::GET_DISPLAYNAME
-			| Backend::SET_DISPLAYNAME // not really, but it is stored in the account table
 			| Backend::PROVIDE_AVATAR
 			| Backend::COUNT_USERS)
 			& $actions);
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -215,10 +215,6 @@ class User_Proxy extends Proxy implements IUserBackend, UserInterface, IProvides
 		return $this->handleRequest($uid, 'getDisplayName', array($uid));
 	}
 
-	public function setDisplayName($uid, $displayName) {
-		// ignored, function needed to make core store the display name in the account table
-
-	}
 	/**
 	 * checks whether the user is allowed to change his avatar in ownCloud
 	 * @param string $uid the ownCloud user name


jvillafanez:
Working in OC 10.0.7 + this branch.

tomneedham:
@butonic why did we have to add this in....?

tomneedham:
I think this might not be required with the sync changes in 10.0.8. I think it was an old bug that wouldnt sync the displayname unless this was set. Need to recheck

jvillafanez:
This is needed to bring back the same behaviour we have in 9.1.7. The problem isn't about syncing, it's about what are you editing there. Are you editing the user backend, or the account information? What will happen during the next sync if such value has been modified?

This is a quick fix in order to have the same behaviour we had before. There are a lot of discussions pending to move forward with a better and cleaner solution.

jvillafanez:
Rebased

jvillafanez:
bump!

Do we still want this?

PVince81:
Need careful review to make sure there are no side effects related to account table @tomneedham @DeepDiver1975

butonic:
the line was introduced with https://github.com/owncloud/user_ldap/pull/163/commits/e1b8f6d2087bff62982efb995ac1a8fb98ec360e because of syncing, see reason in https://github.com/owncloud/core/pull/29669#issue-154420930

provisioning ldap users via saml was broken, because the user_shibboleth app was trying to set the displayname via the user object: https://github.com/owncloud/user_shibboleth/pull/207/files#diff-eec818f31669ac6d6b4d1161cc38794bR447

at the time the immediate fix was to release a new user_ldap and user_shibboleth app. now the logic sync logic has moved to core and we can get rid of this. but two more things need to be done:
- [ ] kill https://github.com/owncloud/user_ldap/blob/26de89ee7ade0164f504ecb558d3272806eac013/lib/User_Proxy.php#L218-L221 with :fire: 
- [x] increase min oc version to 10.0.8, which at least for master already is the case

jvillafanez:
killed dead code and rebased

tomneedham:
Concerned this is just a revert of https://github.com/owncloud/user_ldap/commit/c88e91c530d164b2567f074162a27c3b0fd07798#diff-4dae59834739e07c67b0df543693a485

tomneedham:
Ah, probably ok, since we now added a check in the sync to only update the backend if set is supported https://github.com/owncloud/core/commit/05a82b590300a8e08808895fae77dd1e078a5d15

This was not the case previously and so it was resolved by adding this hack

PR #221, opened 2018-04-04
Run CI

phil-davis:
and see what goes wrong


phil-davis:
Fail is due to core having sharing autocompletion tests that now use different/longer group names. See #222 for proposed fix to user_ldap.

PR #227, opened 2018-04-13
delete unneeded mailhog url

individual-it:
we don't need that mailhogUrl setting any-more
see owncloud/core#31079

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -30,7 +30,6 @@ default:
             adminUsername: admin
             adminPassword: admin
             regularUserPassword: 123456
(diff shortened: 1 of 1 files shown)


individual-it:
CI fails because backport of change in core was only merged this morning, should work tomorrow

tomneedham:
planned - because we want to keep the release really small and focused with just the sync changes

phil-davis:
@tomneedham this is essential to make the CI pass because of existing acceptance test infrastructure changes in core. It is not a change to any released code.

tomneedham:
Cool - sorry! You guys decide on this and merge as necessary :)

PR #236, opened 2018-04-19
Include an exact match of the search term for the uuid attribute

jvillafanez:
Fix user:sync by allowing searching by the uuid (exact match)

PR #238, opened 2018-04-20
Include an exact match of the search term for the uuid attribute

jvillafanez:
Fix user:sync by allowing searching by the uuid (exact match)

Note of possible problems if the same attribute is used in the "user search attributes" checkbox

Superseds https://github.com/owncloud/user_ldap/pull/236

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1381,9 +1381,21 @@ class Access implements IUserTools {
 	 * @return string the final filter part to use in LDAP searches
 	 */
 	public function getFilterPartForUserSearch($search) {
-		return $this->getFilterPartForSearch($search,
-			$this->connection->ldapAttributesForUserSearch,
-			$this->connection->ldapUserDisplayName);
+		$this->detectUuidAttribute('', true);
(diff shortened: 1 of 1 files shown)


tomneedham:
Motivation / use case here? Is this to improve usability of the single user sync? The admin can add the UUID attribute to the user search attributes to achieve this functionality.

jvillafanez:
I think it's to improve the usability. I'd expect that if I use the -u option to sync only one user by uid, it's should work by default. I find pretty weird that I can't sync one user by uid using the default options, specially when the option clearly specify to use the user id.

PVince81:
discuss and decide whether to keep or close

butonic:
I like the idea, however it adds 'magic' and IMO the proper solution is to properly distinguish username frome userid in core (which we actually started with https://github.com/owncloud/core/pull/32587/files ... it is just that nobody has realized we are doing it as yet another hack)

We should add apis that now allow median searching or exact searching. might be a parameter or a new function.

close and implement https://github.com/owncloud/core/issues/29503#issuecomment-462800169 That will create new apis, then we can revisit this properly.

PR #241, opened 2018-04-30
put UI tests into suites

individual-it:
the ui test script now requires the suite parameter. so we place all tests in suites


phil-davis:
Are there other apps that have webUI tests?
We fixed this in ``files_texteditor`` already, and ``firewall`` already has a "proper" suite (I checked last week).

individual-it:
user_managment and notifications, both should be fine

phil-davis:
https://drone.owncloud.com/owncloud/user_ldap/337/103
```
ERROR: webUI tests must be run by suite.
```
It needs something somewhere to pass through suite name(s)

PR #251, opened 2018-06-17
add phan as static code anaylser

patrickjahns:
## Motivation

With https://github.com/owncloud/core/pull/31807 it was a POC for using phan in core for detecting bugs / problems

This PR introduces phan with a lenient configuration to an app and shows how it needs to be configured for testing.

Once this is merged, the next steps include in making the checking stricter and adressing the issues found by phan

Edit: note that i've added quite a few people as reviewers to draw attention and discuss it

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * This configuration will be read and overlaid on top of the
+ * default configuration. Command line arguments will be applied
+ * after this file is read.
+ */
+
+return [
+
+	// Supported values: '7.0', '7.1', '7.2', null.
+	// If this is set to null,
+	// then Phan assumes the PHP version which is closest to the minor version
+	// of the php executable used to execute phan.
+	'target_php_version' => null,
+
+	// A list of directories that should be parsed for class and
+	// method information. After excluding the directories
+	// defined in exclude_analysis_directory_list, the remaining
+	// files will be statically analyzed for errors.
+	//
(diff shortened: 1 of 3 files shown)


phil-davis:
This sort of thing seems good to me. The settings must be very lenient, since you did not have to change any code to make it pass! It will be nice to enforce all sorts of crud that is easily missed if you ignore your IDE - unused vars, undefined vars, missing or wrong phpdoc...

patrickjahns:
@phil-davis  - it's the most lenient setting - same here for core: https://github.com/owncloud/core/pull/31807  - where a lot of bad imports / missing imports have been caught.

increasing the strength - I find already 100 errors for this app - so we need to gradually work here. 
First introduce - ✅ - strengthen checks - fix 💪

jvillafanez:
Nothing against this.

.phan/config.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+/**
+ * This configuration will be read and overlaid on top of the
+ * default configuration. Command line arguments will be applied
+ * after this file is read.
+ */
+
+return [
```
PVince81: should we put this in a common location like the coding standards repo ?

patrickjahns: I would not suggest this - as these settings are depending on the layout of the app (where will code be). And the settings are definitely not the same for every app. Some apps we can have very strict settings fast, whereas some apps will require more lenient settings by design

tomneedham:
👍

PR #253, opened 2018-06-21
Do not log warnings for cache misses

cdamken:
Users are confused that they get this warning for too many users.


lib/User/Manager.php
```diff
@@ -241,7 +241,7 @@ public function getCachedEntry($uid) {
 		}
 
 		// should have been cached during login or while iterating over multiple users, eg during sync
-		$this->logger->warning('No cached ldap result found for '. $uid , ['app' => self::class]);
+		$this->logger->info('No cached ldap result found for '. $uid , ['app' => self::class]);
```
patrickjahns: Agree that `warning` is a bit to harsh in that case? - but why `info` and not `debug` ? cc @butonic

tomneedham:
It's more interesting to find out the cases where this message appears, as the comment suggests, the user should already be cached. Once we have considered that scenario, I think we can evaluate if this should be `info` or `debug`

PVince81:
decide what to do. close ?

butonic:
Log an exception to get the trace. This should not happen because either the user logged in during this request and the first ldap request should have cached his ldap entry or this happens during sync.

I ran into several ussies with multiple basedns, which might lead to entries being pushed out of the cache. the default page size is 500. The default cache size is 512. so in general this should work. but sometimes the ldap search returns more than 512 users even with a page size of 500 ... maybe ... to find out we need more data. so log an exception so we can get our hands on the trace.

PVince81:
@butonic log as debug then ?

butonic:
We should log an error because we should never hit that line. If we still hit it I consider it a bug. Maybe my comment in https://github.com/owncloud/user_ldap/pull/253/files#diff-9a63dbcca373f6250bc3efc1fc92330eR243 did not make that clear enough.

As a fallback we can try to find the user in ldap again. So

- [ ] log an exception with a message indicating where to do a bug report
- [ ] look up the user in ldap if this line is reached

tomneedham:
Damn - this already got added here https://github.com/owncloud/user_ldap/commit/4a876d8dafb80691897b41c26f701f2a1350d91d#diff-9a63dbcca373f6250bc3efc1fc92330eR243

We wanted to know of cases where we expected a cached result, but it didnt exist. Since dynamically fetching a single result from LDAP is what quickly kills performance when not spotted. 

Since this warning is now below these lines, it is porbably never triggered, and the above fetching is happening without us knowing. 

I will change this PR to add the exception above the request to LDAP - and we will see what data we can learn from.

PR #254, opened 2018-06-29
add samaccountname to session

butonic:
use samaccountname for wnd


lib/User_LDAP.php
```diff
@@ -152,6 +152,13 @@ public function checkPassword($uid, $password) {
 		if ($this->config->getSystemValue('enable_avatars', true) === true) {
 			$this->userManager->registerAvatarHook($userEntry);
 		}
+
+		// hack to store a samaccountname in the session
+		$samaccountname = $userEntry->getSAMAccountName();
+		if ($samaccountname !== null) {
+			\OC::$server->getSession()->set('samaccountname', $samaccountname);
```
PVince81: I'd like to call this "alternateLogin" or something more generic to make it reusable

@DeepDiver1975 are you ok with this approach ?
also see https://github.com/owncloud/windows_network_drive/pull/175 and https://github.com/owncloud/core/pull/31951 for usage

if we agree I'll adjust the PRs to rename the attributes

PVince81: had a chat with @DeepDiver1975 and he's ok with this

PVince81:
> Could be here possible to add a configuration option in the config.php like:

in that case I'd rather have a more generic setting called something like `ldap.altloginname_attribute => "cn"` which would then tell the LDAP app to take the altloginname value (which will be used by WND) from another attribute

PVince81:
- [ ] make LDAP attribute configurable that will be used for "altloginname"

butonic:
In the context of ldap this is not an alternate login name. It is the `username`. The `username` is what is used to mount the wnd. In general we need to map several attributes to an owncloud account:
1. `uid`: should be a uuid attribute in ldap, AD uses `objectguid`, openldap uses `entryuuid`. This is used to uniquely identify a user in subsequent sync runs. The user will be looked up by this id and the other attributes will be updated.
2. `displayname`: typically `displayname` in ldap
3. `email`: typically `mail`
4. additional search attributes that are synced and allow finding the user in the sharing dialog.

Currently ownCloud cannot explicitly store the username that might have been configured as the login attribute, eg `cn` or `samaccountname`. Basically it gets lost after login. 

To clarify: there is no alternate login. It literally **IS** the `username`.

I'll clean this up later. Currently looking into the core part.

PVince81:
then we could rename "altLoginName" in the core PRs to something like "extusername" or "externalusername" ? (meaning the user name from the external system, as it's not stored currently)

lib/User_Proxy.php
```diff
@@ -32,9 +32,20 @@
...
+class User_Proxy
+	extends
+		Proxy
+	implements
+		IUserBackend,
+		UserInterface,
+		IProvidesQuotaBackend,
+		IProvidesExtendedSearchBackend,
+		IProvidesEMailBackend,
+		IProvidesUserNameBackend
```
PVince81: since this only exists in 10.0.10 through https://github.com/owncloud/core/pull/32579/, we need to set min-version to 10.0.10

PVince81: @butonic please adjust

butonic:
I renamed alt login name to username because that is what it is. `ldapUserName` is currently redundant with the `ldapExpertUsernameAttr` but the latter determines the uid that will be used in owncloud. we want to move to uuid there anyway. but that is a different issue. For now, I want to name the new option correctly.

It can currently only be set via cli and defaults to `samaccountname`. in core it is only used for wnd mounts, so no harm done.

PVince81:
some test needs adjusting for the new attribute:
```
' contains "| ldapAgentPassword             |              |".

/var/www/owncloud/apps/user_ldap/tests/unit/CommandTest.php:67
```

PR #256, opened 2018-07-04
Restore avatar behaviour of OC 9.1

jvillafanez:
Fix https://github.com/owncloud/enterprise/issues/2553

Checked against OC 10.0.9RC2. There shouldn't be any extra core requirement.

### DIFF
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -240,6 +240,11 @@ class Manager {
 			return $this->usersByUid[$uid];
 		}
 
+		$dn = $this->username2dn($uid);
+		if ($dn) {
+			return $this->getUserEntryByDn($dn);
+		}
+
 		// should have been cached during login or while iterating over multiple users, eg during sync
 		$this->logger->warning('No cached ldap result found for '. $uid , ['app' => self::class]);
 		return null;
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -240,9 +240,16 @@ class Manager {
 			return $this->usersByUid[$uid];
 		}
 
-		$dn = $this->username2dn($uid);
-		if ($dn) {
-			return $this->getUserEntryByDn($dn);
+		try {
+			$dn = $this->username2dn($uid);
+			if ($dn) {
+				return $this->getUserEntryByDn($dn);
+			}
+		} catch (DoesNotExistOnLDAPException $e) {
+			return null;
+		} catch (\OutOfBoundsException $e) {
+			// dn might be outside of the base
+			return null;
 		}
 
 		// should have been cached during login or while iterating over multiple users, eg during sync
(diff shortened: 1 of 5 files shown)


PVince81:
unit tests ?

jvillafanez:
rebased

PVince81:
@tomneedham can you review this ?

lib/User/Manager.php
```diff
@@ -240,6 +240,18 @@ public function getCachedEntry($uid) {
...
+			$dn = $this->username2dn($uid);
+			if ($dn) {
+				return $this->getUserEntryByDn($dn);
+			}
+		} catch (DoesNotExistOnLDAPException $e) {
+			return null;
+		} catch (\OutOfBoundsException $e) {
+			// dn might be outside of the base
+			return null;
+		}
```
tomneedham: Was this required to restore the avatar behaviour? It should just be the below line that is required. The reason I am concerned with this change is that when we optimised the sync code, it should not need to do single entry retrievals from LDAP, they should be cached when loading pages of results. Hence we did not need to load them manually here. What was the reason for the change here @jvillafanez ?

jvillafanez: Yes, it's required. I don't remember the exact use case, but I'm pretty sure that without this change it crashes because the entry isn't cached yet.
Maybe https://github.com/owncloud/core/blob/5a9909dde20c753c6a4c018e8dba83012ed08756/settings/Panels/Personal/Profile.php#L101 . You have to check the backend there in order to check if the user has an avatar set or not in the LDAP server.

jvillafanez: Maybe it doesn't crashes, but it doesn't behave according to the expectations. Quoting the original ticket:
> A user without avatar in LDAP can add his own avatar.
A user with an avatar in LDAP cannot set his own avatar.

If the user isn't cached as it seems, the LDAP user won't be able to set his avatar in any circumstance.

tomneedham:
With this PR, the avatar button appear after I delete the avatar on LDAP, yet they are not clickable...? 

screen shot 2018-08-15 at 12 09 35

tomneedham:
screen shot 2018-08-15 at 12 10 41
 It appears that the language form overlaps, only the top few pixels of the button are clickable... Anyway, not sure this is an issue with this PR - side effect

jvillafanez:
Buttons are clickable with OC 10.0.9 + this branch.

> It appears that the language form overlaps, only the top few pixels of the button are clickable... Anyway, not sure this is an issue with this PR - side effect

It's a css problem. The used height by the non-editable display name and email is less than before, which causes the right side of the profile section to shrink enough to move the Language section (which is part of the profile) to the right.

As far as I know, this is controlled by core, so nothing to do in the app. To be decided how we can solve this css issue.

tomneedham:
Yes agreed - core css problem anyway.

PR #262, opened 2018-07-26
Use acceptance test run.sh for webUI tests

phil-davis:
After core https://github.com/owncloud/core/pull/31348 it is possible to run webUI acceptance tests from the common ``tests/acceptance/run.sh`` script, so do it.


phil-davis:
Needs the nightly update of core image, then run CI again.

PR #264, opened 2018-08-07
always use occ by testing app

individual-it:
after https://github.com/owncloud/core/pull/32239 we can always use the remote version of occ by the testing app


tests/acceptance/features/bootstrap/UserLdapGeneralContext.php
```diff
@@ -393,7 +393,7 @@ public function connectToLdap($suiteParameters) {
 		$this->ldap->bind();
 		SetupHelper::init(
 			"admin", (string)$suiteParameters['adminPassword'],
```
phil-davis: This whole call should be like:
```
		SetupHelper::init(
			$this->featureContext->getAdminUsername(),
			$this->featureContext->getAdminPassword(),
			$this->featureContext->getBaseUrl(),
			$this->featureContext->getOcPath()
		);
```
and you can also remove ``adminUsername`` and ``adminPassword`` from the suite parameters in ``behat.yml``

PR #269, opened 2018-08-17
Dont throw exception when user not found on ldap during login

tomneedham:
When we try to login, we do a lookup on the LDAP server. Unfortunately when not found it just threw a basic exception which could be the same as connection error etc. So I swapped it to use the proper DoesNotExistOnLDAPException and catch this and not log it, but still log other exceptions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /build/
 /vendor/
 .php_cs.cache
+/tests/output/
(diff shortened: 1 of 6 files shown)


PVince81:
@tomneedham please fix the conflicts

tomneedham:
rebased and fixed conflicts

patrickjahns:
@ownclouders rebase

tomneedham:
screen shot 2018-09-06 at 12 06 27
guess I'll do the rebase

patrickjahns:
Shall we merge this for the next release - saw this flooding a lot of logs

PR #276, opened 2018-08-27
su-exec www-data is not needed for acceptance tests

phil-davis:


phil-davis:
Problems with local storage moves. That will need to be looked into to see if something is happening that depends on being "local" in the back-end of the server.

phil-davis:
See #284 
The extra code in this PR related to ``local_storage`` is no longer needed.

PR #278, opened 2018-08-29
Trigger CI

phil-davis:


phil-davis:
Fails creating an emoji in a comment.
Issue raised: #279

Issue #279, opened 2018-08-29
Cannot add a comment containing an emoji when an LDAP user


phil-davis:
CI fails for:
```
--- Failed scenarios:

/var/www/owncloud/tests/acceptance/features/apiMetadataApps/comments.feature:16
``
which is testing an emoji in a comment.

Example drone run: https://drone.owncloud.com/owncloud/user_ldap/576/206

phil-davis:
Adding QA team label so someone ( @individual-it ?) can first have a look at the test and confirm there is a real problem.

individual-it:
works fine in a local test (automated test as well as manual test) - digging

individual-it:
does not work with local drone run when using mysql but works when using sqlite
looking at the `config.php` it does not have `'mysql.utf8mb4' => true,` is the database in the docker container not setup correctly? @patrickjahns ideas?

patrickjahns:
It's `image: mysql:5.5` - in core we use mysql:5.7 for utf8mb4 - ref: https://github.com/owncloud/core/blob/master/.drone.yml#L420-L422

phil-davis:
Fixed by #281

PR #280, opened 2018-08-29
Php72 drone

davitol:
add PHP versions for unit tests


patrickjahns:
Failure related to #279

patrickjahns:
Please squash the commits

patrickjahns:
@davitol - please rebase against current master - we should be green then

PR #288, opened 2018-09-29
Update Behat contexts needed for core acceptance tests

phil-davis:
and while I am here, sort the lists of contexts into alphabetical order to make it easier to compare to core.


phil-davis:
@individual-it this is passing, so it would be good to get it merged so that we have passing CI.

Then I can refactor again to reduce the places where literal passwords are repeated.

PR #289, opened 2018-09-30
Update behat yml 20180930

phil-davis:
and while I am here, sort the lists of contexts into alphabetical order to make it easier to compare to core.


phil-davis:
This is not needed

PR #307, opened 2018-11-12
Disable parallel paginated search for multi base dn on subsequent pages

butonic:
When multiple base dns are configured the `AllUsersIterator` uses pages of 500 to page through the ldap results. It literally provides limit and offset:

https://github.com/owncloud/core/blob/d2697818a682904a13ba8d18e96411722853dd8a/lib/private/User/Sync/AllUsersIterator.php#L40-L47

That confuses the internal parralel seach cookie handling:
1. the first base dn tree only contains ... say 3 results.
2. the second base dn tree contains 1000
the first one will have no further results after the first page, causing the search to be abandoned -> which means all cookies are removed. even for the second base dn tree.

while that shouldn't matter, the next page starts with offset 500.

again, the first base dn tree has no results, abandoning the search and deleting all cookies ... so the sync has no chance of iterating over all pages.

something like that ... It is not 100% correct because it is hard to follow the code path. I need to build a test to cover this automatically.

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -736,7 +736,7 @@ class Access implements IUserTools {
 		$this->connection->writeToCache("userExists$ocName", true);
 	}
 
-	/*
+	/**
 	 * fetches a list of users according to a provided loginName and utilizing
 	 * the login filter.
 	 *
@@ -766,6 +766,13 @@ class Access implements IUserTools {
 	}
 
 	/**
+	 * fetches a list of users according to a provided filter and attributes
+	 *
+	 * WARNING: Using this function combined with LIMIT $limit and OFFSET $offset
+	 * will search in parallel all provided base DNs in this server,
+	 * and thus can return more then LIMIT $limit users. This function shall
+	 * be used with limit and offset by iterators that can
+	 * support this kind of parallel paging.
 	 *
 	 * @param string $filter
 	 * @param string|string[] $attr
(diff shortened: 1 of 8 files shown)


PVince81:
no reviewers, assuming still WIP ?

PVince81:
@butonic please specify what needs to be done to finish this so we can take over

micbar:
@butonic Could you rebase and clarify, what needs to be done?

micbar:
Please rebase and merge.

Seems that this is working. Tested by customers.

micbar:
Is Access.php Unit testable?

VicDeo:
@micbar 
> Is Access.php Unit testable?

I'm afraid it's too complex.

mrow4a:
@micbar @VicDeo closing -> Detected while writing integration tests that this fix is conceptually wrong. Lets assume we request users with limit 2 and offset 2. With this fix, and e.g. 3 base DNs, there is possibility that we will return 3x2users, 2 for each base DN, thus exceeding limit of 2!

mrow4a:
we discussed with @butonic and this is expected behaviour -> add comments in the code and retest.

mrow4a:
@micbar @jvillafanez ready for rereview -> I am pretty confident on paging in LDAP with this PR merged. 

Tested with integration tests, but also manually:
- in UI counting with multiple base dns in various configurations
- syncing users, also with changing user attributes in ldap server and sync
- searching for users

jvillafanez (approved):
We might need some additional comments in the search functions to warn that it's possible to get results over the limit. Other than that, the code looks fine.

mrow4a:
@jvillafanez will add the comments asap

jvillafanez (approved):

PR #315, opened 2018-11-26
fix paging when limit is used

butonic:
the existing pagination code works correctly limit is 0. but when it is >0 it assumes that pagination should be disabled and always abandons all paged queries (which triggers the paging size to be set to 0 and immediately afterwards to the configured page size) This resets paged searches for eg edirectory netiq.

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1053,8 +1053,9 @@ class Access implements IUserTools {
 			$cr = $this->connection->getConnectionResource();
 			foreach ($sr as $key => $res) {
 				if ($this->getLDAP()->controlPagedResultResponse($cr, $res, $cookie, $estimated)) {
+					$range = $offset . "-" . ($offset + $limit);
 					\OC::$server->getLogger()->debug(
-						'Page response cookie '.$this->cookie2str($cookie)." at $limit/$offset, estimated<$estimated>",
+						'Page response cookie '.$this->cookie2str($cookie)." at $range, estimated<$estimated>",
 						['app' => 'user_ldap']);
 					$this->setPagedResultCookie($base[$key], $filter, $limit, $offset, $cookie);
 				}
@@ -1935,15 +1936,16 @@ class Access implements IUserTools {
 		$pagedSearchOK = false;
 		if ($this->connection->hasPagedResultSupport && ($limit !== 0)) {
 			$offset = (int)$offset; //can be null
+			$range = $offset . "-" . ($offset + $limit);
 			\OC::$server->getLogger()->debug(
 				"initializing paged search for  Filter $filter base ".\print_r($bases, true)
-				.' attr '.\print_r($attr, true)." at $limit/$offset",
+				.' attr '.\print_r($attr, true)." at $range",
 				['app' => 'user_ldap']);
 			//get the cookie from the search for the previous search, required by LDAP
 			foreach ($bases as $base) {
 				$cookie = $this->getPagedResultCookie($base, $filter, $limit, $offset);
 				\OC::$server->getLogger()->debug(
-					"Cookie for $base at $limit/$offset is ".$this->cookie2str($cookie),
+					"Cookie for $base at $range is ".$this->cookie2str($cookie),
 					['app' => 'user_ldap']);
 				if (empty($cookie) && $cookie !== '0' && ($offset > 0)) {
 					// no cookie known, although the offset is not 0. Maybe cache run out. We need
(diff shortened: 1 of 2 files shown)


butonic:
cc @micbar

butonic:
maybe related https://github.com/owncloud/core/issues/17173

micbar:
Current feedback from real test:
Sync doesn't stop.

This is consistent with the test results from drone. The integration Test does not finish and runs into php_max_execution time (3600sec)

micbar:
See https://github.com/owncloud/enterprise/issues/3014#issuecomment-442072475

micbar:
Due to this code change, we are able to deactivate paged search
owncloud/user_ldap@df3ac30

This works now with ldapPagingSize 0
and avoids the known bug in eDirectory, which has been fixed in 9.1.1

PVince81:
@butonic is this conceptually acceptable or needs further tweaks ?

asking this to be able to take over and finish

butonic:
I'll look into this tomorrow

lib/Access.php
```diff
@@ -1086,7 +1086,7 @@ private function processPagedSearchStatus($sr, $filter, $base, $iFoundItems, $li
 			foreach ($sr as $key => $res) {
 				if ($this->getLDAP()->controlPagedResultResponse($cr, $res, $cookie, $estimated)) {
 					\OC::$server->getLogger()->debug(
-						'Page response cookie '.$this->cookie2str($cookie).", estimated<$estimated>",
+						'Page response cookie '.$this->cookie2str($cookie)." at $limit/$offset, estimated<$estimated>",
```
jvillafanez: `at $offset/$limit` ? 500/2000 seems better.

lib/Access.php
```diff
@@ -1258,6 +1258,17 @@ private function search($filter, $base, $attr = null, $limit = null, $offset = n
 			return [];
 		}
 
+		unset($findings['count']);
+
+		//we slice the findings, when
+		//a) paged search unsuccessful, though attempted
+		//b) no paged search, but limit set
```
jvillafanez: Does paged search with multiple servers work properly? I think they return more results than the limit, so we might still need to slice in that case too.
I think it's easier if we just check if the number of entries are above the limit, not sure how difficult it would be.

jvillafanez:
I assume that multiple servers are out of scope for this PR. https://github.com/owncloud/user_ldap/blob/deb26568c9f710d96b1550dc99354d57c6c691a1/lib/Access.php#L1264-L1274 is untouched in this PR, but I don't know if it will affect the result.

jvillafanez:
Not sure what is left to do here other than updating the branch. I don't think we can add test here with the current code.

jvillafanez:
rebased

lib/Access.php
```diff
@@ -1964,20 +1972,26 @@ private function initPagedSearch($filter, $bases, $attr, $limit, $offset) {
 					}
 				}
 				if ($cookie !== null) {
-					//since offset = 0, this is a new search. We abandon other searches that might be ongoing.
-					$this->abandonPagedSearch();
```
jvillafanez: There is an infinite loop in the "tests/integration/Lib/IntegrationTestPaging.php" caused by these changes. An unconditionally `abandonPagedSearch` fixes the problem.

PVince81: is there another condition that we can use to break the loop ?

jvillafanez: https://github.com/owncloud/user_ldap/pull/315/commits/9ed66eb3ecff1740cc622bcee70910510dd8024b should fix it, which solves part of the problems in https://github.com/owncloud/core/issues/6311

jvillafanez:
Maybe @butonic wants to double-check my changes before merging.
Unittests can't be added at the moment, so coverage can't be increased.

lib/Access.php
butonic: the order is always limit then offset. in all api calls, in SQL it is `SELECT ... LIMIT 10 OFFSET 1000` ...

butonic:
It should be possible to mock the ldap connection wrapper and return the necessary paging results. It would take me the better part of a day to debug this ... maybe @micbar can get hold of a diagnostics run at an instance that faces this problem, so we can reproduce the ldap responses in unit tests?

jvillafanez:
Screen Shot 2019-03-12 at 17 17 52

Just saying that the recommended maximum CRAP index is usually 30. I wouldn't mind to have some classes over that maximum as long as it's justified.
If you manage to reduce the CRAP index of the Access class to around 30 just with unittests, you deserve a rise.

My point here is that we should start considering to rewrite this part of the code and make it testable instead of spending quite a lot of time trying to make unittests here. I'm not saying is imposible, but if the goal is to spend 1 day to add 1 unittest for 1 case, something is wrong.

lib/Access.php
jvillafanez: I'd bet there are a lot of people that just by reading "Page response cookie 0x00000 at 50/300" won't know what you're talking about. You HAVE TO know that the first number is the limit and the second is the offset, but if you don't know that most of the people will think that we're on point 50 of a maximum of 300.
I changed this to a range, like "Page response cookie 0x00000 at 300-350" which is more natural to read.

PVince81:
in regards to the above comments about unit test complexity and refactoring I'll merge this branch tomorrow unless there are objections @butonic @jvillafanez

PR #318, opened 2018-11-29
Trigger CI

phil-davis:
to see if we have a repeatable CI issue.


phil-davis:
An ordinary PR passes (at least this time)

PR #323, opened 2018-12-06
Release 0.13.0 changelog + VCS tweaks

PVince81:


PVince81:
the package-lock was always there, generated but someone forgot to check in.

it currently only brings in "bower" which is the old package manager for bower.json...

Issue #292, opened 2018-10-09
Add occ command acceptance tests


phil-davis:
``cli`` acceptance tests now exist in core.

General QA issue https://github.com/owncloud/QA/issues/570

- [x] check that extra non-LDAP local users and groups can be created, modified, deleted. #337 
- [x] check that LDAP-defined users can be deleted locally, but can "come back" again (because they are synced from LDAP) #337
- [x] check that LDAP-defined groups can be deleted locally, but can "come back" again (because they are synced from LDAP) #342 
- [x] check what can/cannot happen with respect to modifying attributes of LDAP users using occ commands #341
(clues in comment https://github.com/owncloud/core/pull/32731#issuecomment-428204236 )

phil-davis:
Also review what Provisioning API and webUI acceptance tests for LDAP-specific cases already exist, and add scenarios to fill out coverage of the items above.

dpakach:
Completed in PRs #337 , #341, #342

PR #336, opened 2019-01-09
Run Core Provisioning tests in drone

dpakach:
Related Issue #292

Run core CLI provisioning tests in User Ldap

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -135,6 +135,16 @@ default:
         - WebUISharingContext:
         - WebUIFilesContext:
 
+    cliProvisioning:
+      paths:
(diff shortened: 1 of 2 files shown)


tests/acceptance/config/behat.yml
```diff
@@ -121,6 +121,14 @@ default:
         - FeatureContext: *common_feature_context_params
         - LoggingContext:
 
+    cliProvisioning:
+      paths:
+        - '%paths.base%/../features/cliProvisioning'
```
phil-davis: It also needs a matrix entry in ``.drone.yml``
And I expect that the path here needs to be like above:
```
%paths.base%/../../../../../tests/acceptance/features/cliProvisioning
```
because it needs to find its way out of the app folders back to core.

dpakach:
Running core tests is not possible so moved to  #337

PR #337, opened 2019-01-11
Add CLI tests for user and groups

dpakach:
Related to issue #292

Add cli tests to run to check that the basic user and group features
work when the user_ldap app is enabled


tests/acceptance/features/bootstrap/UserLdapGeneralContext.php
```diff
@@ -102,6 +102,7 @@ public function ldapConfigHasTheseSettings($configId, TableNode $table) {
 
 	/**
 	 * @Given the LDAP users have been resynced
+	 *@When the LDAP users are resynced
```
skshetry:
```suggestion
	 * @When the LDAP users are resynced
```
Add a space

PR #340, opened 2019-01-15
Use occUsersGroupsContext for cli tests

### LINKED ISSUE #339: Fix behat.yml to use new occ contexts
The occ context in acceptance tests was split into two more contexts in PR https://github.com/owncloud/core/pull/34131 . But this app still uses only occContext.
That needs to be fixed.

dpakach:
Fixes #339

Use `occUsersGroupsContext` in behat.yml instead of `occContext`


dpakach:
> I expect that it will still need OccContext also

Yes, I didnt noticed that earlier. I have fixed that

dpakach:
Already merged in #349

Issue #268, opened 2018-08-16
groups not synced for API calls


individual-it:
1. create a new group in LDAP
2. share with this new group using the sharing API e.g. `curl -u user0:1234 http://localhost/owncloud-core/ocs/v1.php/apps/files_sharing/api/v1/shares -d "path=FOLDER&shareType=1&permissions=31&shareWith=grp6"`

result -> `  <message>Please specify a valid group</message>`

syncing users via `user:sync "OCA\User_LDAP\User_Proxy"` does not help

`./occ ldap:update-group grp6` does not work:
```
Group membership attribute is critical for this command, please verify.
* localhost:389 -> memberUid
The following groups are missing in the DB and will be skipped: grp6
```

**BUT** after `./occ group:list` the group is listed and can be used

individual-it:
might be also the reason for  #229

individual-it:
the issue is not limited to API calls, this issue also causes other tests to fail and only to run in the second iteration https://github.com/owncloud/user_ldap/issues/350

PR #356, opened 2019-01-22
extra group:list to sync group information before sharing

### LINKED ISSUE #350: test "sharing with non unique group name" always fails the first time
This test seems always to fail the first time and 
```
 Scenario: sharing with non unique group name (using unique oC group name)         # /var/www/owncloud/apps/user_ldap/tests/acceptance/features/webUIUserLDAP/groupMembership.feature:44
    Given the administrator creates group "grp1" in ldap OU "TestUsers"             # UserLdapUsersContext::createLdapGroup()
    And the administrator adds user "user3" to group "grp1" in ldap OU "TestUsers"  # UserLdapUsersContext::addUserToLdapGroup()
    When the user shares folder "simple-folder" with group "grp1_2" using the webUI # WebUISharingContext::theUserSharesFileFolderWithGroupUsingTheWebUI()
    When the user re-logs in as "user3" using the webUI                             # WebUILoginContext::theUserRelogsInUsingTheWebUI()
    Then folder "simple-folder (2)" should be listed on the webUI # WebUIFilesContext::fileFolderShouldBeListedOnTheWebUI()
```

individual-it:
run an extra `occ group:list` before sharing so that the groups are correctly synced
see #268 
fixes #350

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -138,6 +138,7 @@ default:
         - WebUIUserContext:
         - WebUISharingContext:
         - WebUIFilesContext:
+        - OccContext:
 
     apiUserLDAPConnection:
       paths:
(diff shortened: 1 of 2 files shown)


phil-davis:
So if you run just that 1 scenario, it will fail. But if you run lots of scenarios together then it passes.
Is that the problem?

individual-it:
if you run only that one scenario locally it would fail always
if the whole suite runs in drone it would always fail the first time and then always succeed in the rerun of failed tests.

PR #361, opened 2019-01-28
test connection to backup server

individual-it:
test if ownCloud works correctly when the main server is unreachable but the backup server is online

### DIFF
--- a/tests/acceptance/features/apiUserLDAPConnection/backupServerConnection.feature
+++ b/tests/acceptance/features/apiUserLDAPConnection/backupServerConnection.feature
@@ -1,0 +1,56 @@
+@api
+Feature: connect to a backup LDAP serer
+
+As an administrator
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiUserLDAPConnection/backupServerConnection.feature
```diff
@@ -0,0 +1,56 @@
+@api
+Feature: connect to a backup LDAP serer
+
+As an administrator
+I want to be able to set up a backup LDAP server
+So that user authentifications still work when the main LDAP server is not reachable
```
phil-davis: s/authentifications still work/authentication still works/

"authentification" is not an English word

individual-it: done

PR #373, opened 2019-02-05
Remove wrongly-named WebUIUsersContext

phil-davis:
When running webUI acceptance tests locally, ``behat`` told me that it did not know about ``WebUIUsersContext``. That is true, because that context in core is called ``WebUIUserContext``

I have no idea why ``behat`` does not complain about this in drone. Anyway the steps in ``WebUIUserContext`` are not needed anyway.

While touching this, sort the contexts so it is easier to look through the list.

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -183,8 +183,8 @@ default:
         - UserLdapGeneralContext:
         - FeatureContext: *common_feature_context_params
         - LoggingContext:
-        - SearchContext:
         - PublicWebDavContext:
+        - SearchContext:
 
     apiWebdavProperties:
       paths:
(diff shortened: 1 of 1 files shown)


phil-davis:
Hmmm - this one needs looking at. In core ``stable10`` there is a ```WebUIUsersContext`` that has the "user_management app" steps. And ``user_ldap`` runs some of those from core ``stable10``

phil-davis:
``user_ldap`` runs various core ``user_management`` webUI tests. In core ``stable10`` ``WebUIUsersContext`` exists and is needed. When running against core ``master`` the ``user_management`` app is loaded and ``WebUIUsersContext`` comes from there.

Closing this PR!

Issue #386, opened 2019-03-12
drone acceptance test failures


phil-davis:
All 3 runs of the `webUIProvisioning` suite failed with:

https://drone.owncloud.com/owncloud/user_ldap/1202/322
https://drone.owncloud.com/owncloud/user_ldap/1202/335
https://drone.owncloud.com/owncloud/user_ldap/1202/522

```
Rerun failed scenario: /var/www/owncloud/server/apps/user_ldap/tests/acceptance/features/webUIProvisioning/groups.feature:16
@webUI @insulated @disablePreviews
Feature: add group
  As an administrator
  I want to be able to add, delete and modify groups via the webUI
  So that I can easily manage groups when user LDAP is enabled

  Background:                                           # /var/www/owncloud/server/apps/user_ldap/tests/acceptance/features/webUIProvisioning/groups.feature:7
    Given user admin has logged in using the webUI      # WebUILoginContext::theUserLogsInUsingTheWebUI()
    And the administrator has browsed to the users page # WebUIUsersContext::theUserBrowsesToTheUsersPage()

  Scenario: Add group with same name as existing ldap group                                                         # /var/www/owncloud/server/apps/user_ldap/tests/acceptance/features/webUIProvisioning/groups.feature:16
    When the administrator adds group "grp1" using the webUI                                                        # WebUIUsersContext::theAdminAddsGroupUsingTheWebUI()
    Then the group name "grp1" should be listed on the webUI                                                        # WebUIUsersContext::theGroupNameShouldBeListed()
    And a notification should be displayed on the webUI with the text "Error creating group: Group already exists." # WebUIGeneralContext::aNotificationShouldBeDisplayedOnTheWebUIWithTheText()
      Failed asserting that two strings are equal.
      --- Expected
      +++ Actual
      @@ @@
      -'Error creating group: Group already exists.'
      +''
SCENARIO RESULT: (fail)

--- Failed scenarios:

    /var/www/owncloud/server/apps/user_ldap/tests/acceptance/features/webUIProvisioning/groups.feature:16

1 scenario (1 failed)
5 steps (4 passed, 1 failed)
0m19.60s (22.17Mb)
webUI test rerun failed with exit status: 1
runsh: Exit code of main run: 1
Makefile:184: recipe for target 'test-acceptance-ldap-webui' failed
make: *** [test-acceptance-ldap-webui] Error 1 
```

Needs investigation to see why this `user_ldap` test started failing.

ownclouders:
[...]

PR #387, opened 2019-03-12
Check for webUI notification earlier in groups.feature:16

phil-davis:
Issue #386 

This might help the notification to be seen.


phil-davis:
This issue is actually that the scenario browses to the users page and sometimes the test code does not wait enough for the page to be ready.

PR #388, opened 2019-03-13
Drop PHP 5.6

phil-davis:
1) core stable10 no longer supports PHP 5.6
https://github.com/owncloud/core/pull/34698

2) core added `WebUIPersonalSharingSettingsContext` and steps it it are now needed for `webUICore` suite. So add that to `behat.yml`

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -358,7 +358,7 @@ matrix:
     - PHP_VERSION: 7.2
       TEST_SUITE: owncloud-coding-standard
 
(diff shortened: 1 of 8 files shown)


phil-davis:
Rebased this just now. I expect that some `webUIProvisioning` `groups.feature` tests against core `stable10` will still fail until https://github.com/owncloud/core/pull/34776 gets into the overnight core `stable10` QA tarball.

PR #391, opened 2019-03-14
Debug web ui provisioning groups 2

phil-davis:


phil-davis:
The "magic sleep" after browsing to the users page is what "fixes" this.
Proper waiting is being implemented in https://github.com/owncloud/user_management/pull/162
After that is merged, we should be able to run this without the sleep.

phil-davis:
Debugging done.

PR #392, opened 2019-03-15
Remove excess require-dev dependencies from composer.json

phil-davis:
`phpunit` - nowadays `Makefile` uses the one from core
`php-parallel-lint` - owncloud codestyle gives us effective lint also
`owncloud/coding-standard` is in `vendor-bin` already, it is not needed up in this top-level `composer.json`

I noticed this because I saw "the bot" doing PR #389 

When this PR is merged, "the bot" will discover that it does not need to do anything.


phil-davis:
Note: `user_ldap` still has this intermittent fail against core `stable10`
https://drone.owncloud.com/owncloud/user_ldap/1278/449

Should be fixed when tonight's `stable10` QA tarball has built. Then we run CI again.

PR #396, opened 2019-03-19
Extract housekeeping part from new LDAP wizard PR

VicDeo:
Extracted from https://github.com/owncloud/user_ldap/pull/210 to reduce the amount of LOC and changed files there.
Note: `lib/AppInfo/Application.php` and `appinfo.xml` are not exactly the same with that PR as it has new classes.


appinfo/app.php
```diff
@@ -26,30 +26,8 @@
  *
  */
 
-$helper = new \OCA\User_LDAP\Helper();
-$configPrefixes = $helper->getServerConfigurationPrefixes(true);
+namespace OCA\User_LDAP\AppInfo;
```
jvillafanez: I'd rather not use a namespace here

lib/AppInfo/Application.php
```diff
@@ -0,0 +1,86 @@
...
+use OCA\User_LDAP\User_Proxy;
+
+class Application extends \OCP\AppFramework\App {
+	/**
+	 * @param array $urlParams
+	 */
+	public function __construct($urlParams = []) {
+		parent::__construct('user_ldap', $urlParams);
+		$this->registerService();
+		$this->checkCompatability();
```
jvillafanez: just checking, this is the usual thing to be done once, when the app is enabled. Not sure if there is a better place for this check.
Another option could be to explicitly call this function in the app.php file.

lib/AppInfo/Application.php
```diff
@@ -0,0 +1,86 @@
...
+				['app' => 'user_ldap']);
+		}
+	}
+
+	public function registerBackends() {
+		$helper = new Helper();
+		$configPrefixes = $helper->getServerConfigurationPrefixes(true);
+		if (\count($configPrefixes) > 0) {
+			$ldapWrapper = new LDAP();
+			$ocConfig = \OC::$server->getConfig();
```
jvillafanez: Better pick it from the `$this->getContainer()->getServer()`

lib/AppInfo/Application.php
```diff
@@ -0,0 +1,86 @@
...
+			$userBackend  = new User_Proxy($configPrefixes, $ldapWrapper, $ocConfig);
+			$groupBackend  = new Group_Proxy($configPrefixes, $ldapWrapper);
+			// register user backend
+			\OC_User::useBackend($userBackend);
+			\OC::$server->getGroupManager()->addBackend($groupBackend);
```
jvillafanez: same here

jvillafanez:
Could you also confirm (just to double-check) that this PR can be safely merged without the rest of the stuff that will come?

lib/AppInfo/Application.php
VicDeo: If it will checked only when app is enabled `user_webdavauth` can be enabled after the `user_ldap` and nobody will know that apps are incompatible

VicDeo:
@jvillafanez as you see I moved the current app.php into the Application class.
the same with register_command.php - it is moved into appinfo.xml.
So yes, this is safe.

lib/AppInfo/Application.php
VicDeo: moved into app.php

lib/AppInfo/Application.php
VicDeo: done

appinfo/app.php
VicDeo: removed

lib/AppInfo/Application.php
VicDeo: done

VicDeo:
@jvillafanez everything is addressed, please rereview

PR #397, opened 2019-03-21
Split configuration into config and mapper

VicDeo:
Another step towards https://github.com/owncloud/user_ldap/pull/210
The code that creates/loads/saves/deletes configurations is moved to the separate class


VicDeo:
well, this does not work...
I need to collect all public properties of the Configuration class around the code first.

jvillafanez:
I have some doubts about the new Config class.
* Why do we need to distinguish between "data" and "rawData"? I think it's better to initialize the "data" attribute with the default values and override those values with the new ones, either when they're loaded from the DB or when they're being set one by one. You can store the id in an additional attribute to make the serialization a bit easier

    I've seen a few places where we are still checking whether the configuration is the default one. There are some questions to answer:
    * What are the reasons to keep those checks?
    * Are those reasons still valid?
    * Do we have any case where we **have to** know what are the default values?
    * Is it enough just to know that the configuration has been modified and it isn't the default any longer?
* The constructor for the object is too open. You can use any object as the "data" parameter without any complain.

    I think the easiest solution is to provide a factory to create the Config objects. Standard usage would be something like `$configFactory->getConfig($id)` which would return a Config object with that id. This factory can depend on the `ConfigMapper` to fetch the configuration from the DB. It can also create new Config objects that could be saved through the ConfigMapper (something like `$configFactory->getNewConfig()` with a id fetched from `nextPossibleConfigurationPrefix()`).

    This seems an easier way to create the Config objects. In addition, you can transparently inject the required `\OC::$server->getConfig()` in the Config object (if we can't get rid of it).

VicDeo:
@jvillafanez
>I've seen a few places where we are still checking whether the configuration is the default one. There are some questions to answer:
- What are the reasons to keep those checks?
- Are those reasons still valid?
- Do we have any case where we have to know what are the default values?
- Is it enough just to know that the configuration has been modified and it isn't the default any longer?


`isDefault` is used to validate the LDAP configuration. For the default config validation process is terminated early. This is the only reason to have such method.
It is used in the single place: **Connection.php**
See https://github.com/owncloud/user_ldap/pull/410/files

VicDeo:
> Why do we need to distinguish between "data" and "rawData"? I think it's better to initialize the "data" attribute with the default values and override those values with the new ones, either when they're loaded from the DB or when they're being set one by one. You can store the id in an additional attribute to make the serialization a bit easier

Basically I renamed existing Configuration class into `Config/Config`. But ok, `rawData` has been removed.

jvillafanez:
> isDefault is used to validate the LDAP configuration. For the default config validation process is terminated early. This is the only reason to have such method.
It is used in the single place: Connection.php
See https://github.com/owncloud/user_ldap/pull/410/files

"wait until that is merged" I guess.
Should we merge it now and update this PR?

VicDeo:
@jvillafanez if you're fine with that change, we can do it according to your suggestion.

micbar:
@ownclouders rebase

lib/Config/Config.php
```diff
@@ -0,0 +1,333 @@
...
+	public function getData() {
+		return $this->data;
+	}
+
+	/**
+	 * @param string $name
+	 *
+	 * @return mixed|null
+	 */
+	public function __get($name) {
```
jvillafanez: I guess this is legacy and we can't replace it easily....

lib/Config/Config.php
```diff
@@ -0,0 +1,333 @@
...
+	 * @param array &$applied optional; array where the set fields will be given to
+	 * @return false|null
+	 */
+	public function setConfiguration($config, &$applied = null) {
+		if (!\is_array($config)) {
+			return false;
+		}
+
+		foreach ($config as $inputKey => $val) {
+			if (\array_key_exists($inputKey, $this->data)) {
```
jvillafanez: since `isset` is usually prefered over `array_key_exists`, include a comment saying that this usage is intended because some keys might have "null" as value.

lib/Config/Config.php
```diff
@@ -0,0 +1,333 @@
...
+				$key = $inputKey;
```
jvillafanez: I don't think we need to duplicate / clone the key.

lib/Config/Config.php
```diff
@@ -0,0 +1,333 @@
...
+			}
+			if (\is_array($applied)) {
+				$applied[] = $inputKey;
+			}
+		}
+		return null;
+	}
+
+	public function getDefaults() {
+		$defaultConfig = new self([]);
```
jvillafanez: A `Config` object creating another `Config` object to get its data doesn't look good. I think this is another reason to proceed with the `ConfigFactory` idea: you can easily do the same just by `$configFactory->getFreshConfig()->getData();` or even something like `$configFactory->getFreshConfig()->isEqualTo($config);` if we're interested in comparing configurations.

lib/Config/Config.php
VicDeo: removed

lib/Config/Config.php
VicDeo: added

lib/Config/Config.php
VicDeo: @jvillafanez I'd prefer not to refactor too much as rebasing https://github.com/owncloud/user_ldap/pull/210 on top of this PR will give me enough headache already.

jvillafanez: Ok. Still, I think we should open an issue to refactor and cleanup these things after eveything is merged.

lib/Config/ConfigMapper.php
```diff
@@ -0,0 +1,157 @@
...
+		$key = self::PREFIX . $id;
+		$json = $this->config->getAppValue('user_ldap', $key, null);
+		if ($json === null) {
+			throw new DoesNotExistException("$key does not exist");
+		}
+		$this->config->deleteAppValue('user_ldap', $key);
+	}
+
+	public function nextPossibleConfigurationPrefix() {
+		$qb = \OC::$server->getDatabaseConnection()->getQueryBuilder();
```
jvillafanez: This should be injected

jvillafanez:
One last comment and I think this is good to merge if it works properly and there is nothing else to be done.

lib/Config/ConfigMapper.php
VicDeo: done

PR #399, opened 2019-03-29
Remove unused use statement

VicDeo:


phil-davis:
There are more of these - PHPstorm can automagically remove them. See PR #400

Issue #377, opened 2019-02-12
LDAP 0.14.0


PVince81:
When taking care of a task (and marking as done), please add your name to the task!

### Reason
- [x] Add network timeout https://github.com/owncloud/user_ldap/issues/324 
- [x] Allow avatars to be changed by users if not provided by LDAP - [#188](https://github.com/owncloud/user_ldap/issues/188)
- [x] Remove PHP 5.6 support - [#388](https://github.com/owncloud/user_ldap/issues/388)
- [x] ~LDAP configuration is now stored JSON encoded - [#393](https://github.com/owncloud/user_ldap/issues/393)~ reverted for this release 
- [x] Clean up Application initialization code - [#396](https://github.com/owncloud/user_ldap/issues/396)
- [x] Remove unused use statements - [#399](https://github.com/owncloud/user_ldap/issues/399) [#400](https://github.com/owncloud/user_ldap/issues/400)
- [x] Do not return unavailable users - [#12](https://github.com/owncloud/user_ldap/issues/12)
- [x] Fix paging when limit is used - [#315](https://github.com/owncloud/user_ldap/issues/315)
- [x] Reuse existing LDAP accounts if available https://github.com/owncloud/user_ldap/pull/165
- [x] Replace magic numbers with constants https://github.com/owncloud/user_ldap/pull/435
- [x] Log bind errors https://github.com/owncloud/user_ldap/pull/436
- [x] Simplify connection: Get rid of init method https://github.com/owncloud/user_ldap/pull/437

### QA

- [x] Create **QA Milestone**
- [x] Move items from `development` to `qa` @davitol 
- [x] Create release branch with the format "release-0.14.0" @patrickjahns  
- [x] Create "0.14.0 Testplan" @davitol https://github.com/owncloud/user_ldap/issues/446
- [x] Test and document in testplan => Final "Go" from QA (set 'QA-ready' label) @davitol 

### Documentation

- [ ] Changes in the docs needed?

### Marketing

- [ ] Ping RocketChat channel #marketing
- [x] Check/adapt info.xml stuff
- [ ] Check/adapt screenshot
- [ ] Blog post or other communication if applicable

### Build

All actions to be done on the release branch from here:
- [x] Bump version in info.xml (no version suffix)
- [x] Update CHANGELOG.md - Change CHANGELOG.md 'Unreleased' to '$version' (final release date unknown)

#### Beta/RC

- [x] Create github tag for 'v${version}RC1' on release branch
[...]

PR #405, opened 2019-04-08
Update changelog

PVince81:
The item "Do not return unavailabl users" isn't clear enough but will need more extensive research to clarify.
I've added as checkbox for the release here: https://github.com/owncloud/user_ldap/issues/377#issuecomment-480734453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
+- Add network timeout setting - [#324](https://github.com/owncloud/user_ldap/issues/324)
+
+### Changed
+
+- Allow avatars to be changed by users if not provided by LDAP - [#188](https://github.com/owncloud/user_ldap/issues/188)
+- Remove PHP 5.6 support - [#388](https://github.com/owncloud/user_ldap/issues/388)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+
+- Allow avatars to be changed by users if not provided by LDAP - [#188](https://github.com/owncloud/user_ldap/issues/188)
+- Remove PHP 5.6 support - [#388](https://github.com/owncloud/user_ldap/issues/388)
+- LDAP configuration is now stored JSON encoded - [#393](https://github.com/owncloud/user_ldap/issues/393)
+- Clean up Application initialization code - [#396](https://github.com/owncloud/user_ldap/issues/396)
+- Remove unused use statements - [#399](https://github.com/owncloud/user_ldap/issues/399) [#400](https://github.com/owncloud/user_ldap/issues/400)
+
+### Fixed
+
+- Do not return unavailable users - [#12](https://github.com/owncloud/user_ldap/issues/12)
```
phil-davis: Maybe "only return users valid for ownCloud when getting LDAP group members"
The issue is that there can be a filter in ownCloud that reduces the set of users from the LDAP server that are allowed to use ownCloud. But their might be an LDAP group that has a mix of users that are and are not in the filter for ownCloud. So when the user_ldap app asks for a list of the members in the group, we just want to find the ones that are valid for ownCloud.

PVince81: adjusted. thanks.

I didn't have enough insight about what this one was about

PR #407, opened 2019-04-12
Rename controller args and json keys

VicDeo:
Do some TODOs and fix one FIXME 

Basically renames 
`ldap_serverconfig_chooser` into `id` to match https://github.com/owncloud/user_ldap/pull/210
`ldap_clear_mapping` into `subject`

### DIFF
--- a/lib/Controller/MappingController.php
+++ b/lib/Controller/MappingController.php
@@ -62,12 +62,11 @@ class MappingController extends Controller {
 	/**
 	 * test the given ldap config
 	 *
-	 * @param string $ldap_clear_mapping subject, 'group' or 'user'
-	 * @param string $ldap_serverconfig_chooser config id // FIXME remove in JS, is unneeded
+	 * @param string $subject, 'group' or 'user'
+	 *
 	 * @return DataResponse
 	 */
-	public function clear($ldap_clear_mapping, $ldap_serverconfig_chooser = null) {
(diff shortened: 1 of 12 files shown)


lib/Controller/MappingController.php
```diff
@@ -62,12 +62,11 @@ public function __construct($appName,
...
 	 *
-	 * @param string $ldap_clear_mapping subject, 'group' or 'user'
-	 * @param string $ldap_serverconfig_chooser config id // FIXME remove in JS, is unneeded
+	 * @param string $subject, 'group' or 'user'
+	 *
 	 * @return DataResponse
 	 */
-	public function clear($ldap_clear_mapping, $ldap_serverconfig_chooser = null) {
-		$subject = $ldap_clear_mapping; // TODO if possible make JS send as 'subject' right away
+	public function clear($subject) {
```
jvillafanez: just to confirm the ldap_serverconfig_chooser / id isn't needed (?)

jvillafanez:
Code-wise it looks fine, but I haven't tested it. Maybe @VicDeo can confirm it works without problems.

lib/Controller/MappingController.php
VicDeo: no, it isn't. We truncate the whole user/group table https://github.com/owncloud/user_ldap/blob/master/lib/Mapping/AbstractMapping.php#L259-L264

VicDeo:
@jvillafanez yes, I did a quick UI testing

PR #408, opened 2019-04-15
Fix false negative result when deleting a server configuration

VicDeo:
Discovered while testing https://github.com/owncloud/user_ldap/pull/407
Config deletion didn't remove it from dropdown because deleteServerConfiguration had false negative result.

### DIFF
--- a/lib/Helper.php
+++ b/lib/Helper.php
@@ -110,7 +110,7 @@ class Helper {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


jvillafanez:
It seems the documentation in the ownCloud's query builder is a bit confusing. It might be worthy a PR to update it in order to clarify the behaviour.

PR #409, opened 2019-04-29
Store arrays as arrays, do not use newline as a separator

VicDeo:
This is a followup for https://github.com/owncloud/user_ldap/pull/393
We can encode arrays directly to json and do not use CR/LF as a separator. 

Before
```
"ldapBase":"dc=firstbase,dc=com/ndc=secondbase,dc=com"
```

After 
```
"ldapBase":"[
"dc=firstbase,dc=com",
"dc=secondbase,dc=com"
]
```

Note that the part splitting these values when they are coming from the wizard is still in place.

Additionally, the config options notation has been unified to pascalCase.
E.g. 
before **$configuration->ldap_host** or **$configuration->ldapHost** could be used so the code was a bit messy
This PR removes **$configuration->ldap_host** notation support and makes the frontend responsible for underscore <-> pascalCase transformation.

### DIFF
--- a/appinfo/Migrations/Version20190315192717.php
+++ b/appinfo/Migrations/Version20190315192717.php
@@ -140,6 +140,8 @@ class Version20190315192717 implements ISimpleMigration {
 		'ldap_paging_size'                  => 'ldapPagingSize',
 		'ldap_experienced_admin'            => 'ldapExperiencedAdmin',
 		'ldap_dynamic_group_member_url'     => 'ldapDynamicGroupMemberURL',
+		'ldap_uuid_user_attribute'          => 'ldapUuidUserAttribute',
+		'ldap_uuid_group_attribute'         => 'ldapUuidGroupAttribute'
 	];
 
 	/**
@@ -279,9 +281,10 @@ class Version20190315192717 implements ISimpleMigration {
 
 	/**
 	 * @param  mixed $value
-	 * @return string[]
+	 * @return string | string[]
 	 */
 	private function convertToArray($value) {
+		// try to split the value by all known separators
 		if (empty($value)) {
 			$value = '';
 		} elseif (!\is_array($value)) {
@@ -292,8 +295,10 @@ class Version20190315192717 implements ISimpleMigration {
 		}
 
 		if (!\is_array($value)) {
+			// if the value is not an array - store it as is
 			$finalValue = \trim($value);
 		} else {
+			// if the value is an array - clean all empty values
 			$finalValue = [];
 			foreach ($value as $key => $val) {
 				if (\is_string($val)) {
--- a/appinfo/Migrations/Version20190315192717.php
+++ b/appinfo/Migrations/Version20190315192717.php
@@ -143,6 +143,22 @@ class Version20190315192717 implements ISimpleMigration {
 	];
 
 	/**
+	 * @var string[] - config keys that should be converted into an array
+	 */
+	private $arrayConfigKeys = [
+		'ldapBase',
+		'ldapBaseUsers',
+		'ldapBaseGroups',
+		'ldapAttributesForUserSearch',
+		'ldapAttributesForGroupSearch',
+		'ldapUserFilterObjectclass',
+		'ldapUserFilterGroups',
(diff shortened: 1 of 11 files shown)


appinfo/Migrations/Version20190315192717.php
```diff
@@ -252,9 +268,50 @@ private function getTranslatedLegacyConfig($prefix) {
...
+
 		return $current;
 	}
 
+	/**
+	 * @param  mixed $value
+	 * @return string[]
+	 */
+	private function convertToArray($value) {
+		if (empty($value)) {
```
jvillafanez: maybe include some comments to make it easier to follow?

lib/Configuration.php
```diff
@@ -316,23 +313,17 @@ public function saveConfiguration() {
...
-	 */
-	protected function getMultiLine($varName) {
-		$value = $this->getValue($varName);
-		if (empty($value)) {
-			$value = '';
-		} else {
-			$value = \preg_split('/\r\n|\r|\n/', $value);
+		$c = $this->getTranslatedConfig();
+		// multiple values means the config is not default
+		foreach ($this->arrayConfigKeys as $arrayConfigKey) {
```
jvillafanez: I see some problems with the default values:
* Why the "default" value in the `$data` attribute are different than the ones fetched from the `getDefaults` function? I'd prefer to remove that information and use something like `$this->data = $this->getDefaults()` in the constructor in order to have only one source of information. Then we can overwrite the values accordingly to the specific DB information.
* The `arrayConfigKeys` shouldn't be needed. You have the actual data in `$this->data` and also the default values. Just traverse the array checking each value. Most of the time, the first value (ldapHost) will be different, so it will be a fast check. No need to store an additional array or perform heavy operations.

lib/Configuration.php
```diff
@@ -46,60 +46,19 @@ class Configuration {
...
-		'ldapExpertUsernameAttr' => null,
-		'ldapExpertUUIDUserAttr' => null,
-		'ldapExpertUUIDGroupAttr' => null,
-		'lastJpegPhotoLookup' => null,
-		'ldapNestedGroups' => false,
-		'ldapPagingSize' => null,
-		'ldapDynamicGroupMemberURL' => null,
+	protected $data = [];
+
+	protected $arrayConfigKeys = [
```
jvillafanez: This isn't used any longer

VicDeo: @jvillafanez I know, already deleted in my local copy. I decided to go further and move  underscore_value->pascalCaseValue translation into javascipt

appinfo/Migrations/Version20190315192717.php
VicDeo: done

lib/Configuration.php
VicDeo: removed

lib/Configuration.php
VicDeo: removed

VicDeo:
@jvillafanez please rereview

js/wizard/configModel.js
```diff
@@ -33,6 +33,60 @@ OCA = OCA || {};
 		 * @param {OCA.LDAP.Wizard.WizardDetectorQueue} detectorQueue
 		 */
 		init: function (detectorQueue) {
+			this.translations = {
```
jvillafanez: I'm not sure if this could be confused with language translations...

jvillafanez:
I guess there is no real alternative not to use the translations / mappings in js. Being positive, this could allow us to change the html name to whatever we want.
If there isn't anything we missed along the way, I'm fine with the changes.

VicDeo:
@jvillafanez I can remove these mappings completely but the changeset will be too massive. 
It will affect 20 js file and all templates.
So I just moved it to js as that's enough for pushing https://github.com/owncloud/user_ldap/pull/210  forward - all js and template files anyway will be deleted there anyway.

js/wizard/configModel.js
VicDeo: renamed to `mappings`

PR #415, opened 2019-05-14
run tests on LDAPS

### LINKED ISSUE #348: run acceptance tests with LDAPS
currently all tests run with unencrypted LDAP, we also should run them with LDAPS

individual-it:
fixes  #348

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,14 +6,13 @@ branches: [master, release*]
 
 pipeline:
   ldap:
-    image: dinkel/openldap
+    image: osixia/openldap
     detach: true
     pull: true
     environment:
-      - SLAPD_DOMAIN=owncloud.com
-      - SLAPD_ORGANIZATION=ownCloud
-      - SLAPD_PASSWORD=admin
-      - SLAPD_ADDITIONAL_MODULES=memberof
+      - LDAP_DOMAIN=owncloud.com
+      - LDAP_ORGANISATION=ownCloud
+      - LDAP_ADMIN_PASSWORD=admin
 
   install-core:
     image: owncloudci/core
--- a/.drone.yml
+++ b/.drone.yml
@@ -13,6 +13,8 @@ pipeline:
       - LDAP_DOMAIN=owncloud.com
       - LDAP_ORGANISATION=ownCloud
       - LDAP_ADMIN_PASSWORD=admin
+      - HOSTNAME=ldap
+      - LDAP_TLS_VERIFY_CLIENT=never
 
   install-core:
     image: owncloudci/core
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -126,6 +127,27 @@ pipeline:
...
+      matrix:
+        NEED_SERVER: true
+
+  ldaps-config-for-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /var/www/owncloud/server/
+      - php occ ldap:set-config LDAPTestId ldapPort "636"
+      - php occ ldap:set-config LDAPTestId ldapHost "ldaps://ldap"
```
skshetry: Setting `ldaps://` doesnot require us to set `ldapTLS`?

individual-it: actually both cannot be set at the same time. I had to remove `ldapTLS=1` again

skshetry:
@individual-it There seems to be few tests (~5) that sets host to `ldap`, which are failing.

individual-it:
Yes, when testing failing and reestablished connections we set the host with `ldap://` and the port to `389` that obviously does not work with LDAPS, need something more clever here

tests/acceptance/features/bootstrap/UserLdapGeneralContext.php
```diff
@@ -25,6 +25,7 @@
 use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\TableNode;
 use Behat\MinkExtension\Context\RawMinkContext;
+use PHPUnit\Framework\Assert;
```
skshetry: Do we need to import `Assert` namespace?

tests/acceptance/config/behat.yml
```diff
@@ -8,11 +8,7 @@ default:
         - '%paths.base%/../../../../../tests/acceptance/features'
       context: &common_ldap_suite_context
         parameters:
-          ldapAdminUser: cn=admin,dc=owncloud,dc=com
           ldapAdminPassword: admin
-          ldapBaseDN: dc=owncloud,dc=com
```
skshetry: We do not use this anywhere, do we?

individual-it: that setting is needed for setting the LDAP connection see https://github.com/owncloud/user_ldap/pull/415/files#diff-1fd6e7b9542409bf830dd808e362648aL422
before we used the setting from this yml, now from the current LDAP setting see https://github.com/owncloud/user_ldap/pull/415/files#diff-1fd6e7b9542409bf830dd808e362648aR438

tests/acceptance/features/bootstrap/UserLdapGeneralContext.php
individual-it: good catch, fixed

PR #332, opened 2018-12-21
Simplify and refactor connection establishment

butonic:
When the ldap error certificate changes connection establishment might silently fail. This PR simplifies the code for establishing ldap connections and adds extended diagnostics logging for errors.

PR #437, opened 2019-07-16
Simplify connection: Get rid of init method

VicDeo:
extracted from https://github.com/owncloud/user_ldap/pull/332 
 trying to localize the  part breaking acceptance tests

### DIFF
--- a/lib/Connection.php
+++ b/lib/Connection.php
@@ -165,29 +165,15 @@ class Connection extends LDAPUtility {
 	}
 
 	/**
-	 * initializes the LDAP backend
-	 * @param bool $force read the config settings no matter what
-	 * @throws \OC\ServerNotAvailableException
-	 */
-	public function init($force = false) {
-		$this->readConfiguration($force);
-		$this->establishConnection();
-	}
-
-	/**
 	 * Returns the LDAP handler
+	 * @return resource | null
+	 *
 	 * @throws \OC\ServerNotAvailableException
 	 */
 	public function getConnectionResource() {
-		if (!$this->ldapConnectionRes) {
-			$this->init();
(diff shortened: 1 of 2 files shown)


jvillafanez:
A couple of questions:
* I think there are cases where the connection resource can be null, not a resource. Are those cases properly handled?
* Is it possible to change the connection on-the-fly, or would we need to create a new connection resource? I mean, assuming we have a connection to ldap1, could we change that connection object to point to ldap2, or do we need to create a new connection object? If this isn't possible, I think we should included some comments in whatever method could be risky.

VicDeo:
@jvillafanez 
> I think there are cases where the connection resource can be null, not a resource. Are those cases properly handled?

it is handled by the calling method right now. E.g.
https://github.com/owncloud/user_ldap/blob/master/lib/Connection.php#L634-L639
https://github.com/owncloud/user_ldap/blob/master/lib/Access.php#L178-L184
https://github.com/owncloud/user_ldap/blob/master/lib/Access.php#L958-L965
https://github.com/owncloud/user_ldap/blob/master/lib/Wizard.php#L640-L643
etc.



>Is it possible to change the connection on-the-fly, or would we need to create a new connection resource?

no, new Connection object is required if the configuration has been changed . Like this https://github.com/owncloud/user_ldap/blob/master/lib/Controller/ConfigurationController.php#L138

lib/Connection.php
```diff
@@ -164,30 +164,16 @@ public function setIgnoreValidation($state) {
...
-	 * @throws \OC\ServerNotAvailableException
-	 */
-	public function init($force = false) {
-		$this->readConfiguration($force);
-		$this->establishConnection();
-	}
-
 	/**
 	 * Returns the LDAP handler
+	 * @return resource
```
jvillafanez: include that `null` can also be returned

PR #441, opened 2019-07-27
Add branch specific restrictions to .drone.yml

micbar:
Add new restrictions to .drone.yml

Execute ci on ` master, release*, next* `branches.


patrickjahns:
@micbar 
Keep in mind that you also need to get this restriction into the branch where you want drone to run.
The `drone.yml` is fetched for the current branch / pull request depending on the base

Issue #447, opened 2019-08-06
occ commands throw an exception after upgrading to user_ldap-0.14.0RC1 from an older version


davitol:
### Steps to reproduce

1. Install an oC server using LDAP 0.13.0
2. Log in with an ldap user (checks LDAP works)
3. Via CLI run upgrade ```occ market:install -l user_ldap-0.14.0RC1.tar.gz``` 
4. Run occ app:list command

### Expected behaviour

occ commands should work after the user_ldap upgrade

### Actual behaviour

occ commands fails:

![Screen Shot 2019-08-06 at 10 57 15](https://user-images.githubusercontent.com/10596560/62539080-a8c8b400-b854-11e9-891f-a94543903086.png)

**Note:** Using user_ldap-0.14.0RC1 from scratch (fresh installation) works fine and occ commands are available.

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
PHP 7.2.16-1+ubuntu16.04

**ownCloud version:**

"version":""10.2.1.4" EE"

**Updated from an older ownCloud or fresh install:**
Fresh

**Are you using encryption:** 
No

**Logs**

No Server Logs spotted

```
```

(cc @patrickjahns jfyi)

patrickjahns:
Different szenario which works

- Install owncloud https://download.owncloud.org/community/owncloud-10.2.1.tar.bz2 
- Download userldap from https://github.com/owncloud/user_ldap/releases/tag/v0.14.0RC1
- `occ m:i -l user_ldap-0.14.0RC1.tar.gz`

Everything works

patrickjahns:
Following szenario fails:

- Install owncloud https://download.owncloud.org/community/owncloud-10.2.1.tar.bz2
- Dowload userldap v0.13.0 from https://github.com/owncloud/user_ldap/releases/tag/v0.13.0
- extract contents to `apps/user_ldap`
- `occ app:enable user_ldap`
- Download userldap v0.14.0RC1 from https://github.com/owncloud/user_ldap/releases/tag/v0.14.0RC1
- occ m:i -l user_ldap-0.14.0RC1.tar.gz

Which means we are currently having issues with `apps` and `apps-external`

VicDeo:
introduced by https://github.com/owncloud/user_ldap/commit/360af338bf93bb87a473144a53703ca1e2ff2bc2#diff-224bb875c5d3cfcd7621135069013242

[...]

PR #449, opened 2019-08-19
Hotfix #447

VicDeo:
if we want not to loose compatibility with 10.2

### DIFF
--- a/appinfo/register_command.php
+++ b/appinfo/register_command.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+// FIXME: The command registration should be moved into info.xml once the app compatibility is bumped to OC 10.3
+// see https://github.com/owncloud/user_ldap/issues/447 for details
+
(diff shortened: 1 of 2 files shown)


appinfo/register_command.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
+ * @author Joas Schilling <coding@schilljs.com>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ *
+ * @copyright Copyright (c) 2016, ownCloud GmbH.
```
DeepDiver1975: 2019

appinfo/register_command.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
```
DeepDiver1975: please add correct author

PVince81:
not sure if the hotfix is really needed, considering that the original issue is with the apps/apps_external constellation for which there is a workaround: https://github.com/owncloud/user_ldap/issues/447#issuecomment-521907751

unless we want to send this patch to users and not release / merge it ?

VicDeo:
@DeepDiver1975 
>When creating the command from xml the exact same stuff is done as what has explicitly been coded now.
There is still something wrong if the code patch is not fixing the issue

true. 
but `register_command` is executed significantly later than `info.xml` is parsed. And when it is parsed wrong app.php is loaded. So the real fix is https://github.com/owncloud/core/pull/36054

and a hotfix for the app is partially revert  the changes from https://github.com/owncloud/user_ldap/commit/360af338bf93bb87a473144a53703ca1e2ff2bc2#diff-224bb875c5d3cfcd7621135069013242
This PR does exactly that. The same about authors - I just restored the original `register_command.php`, it had been not written by me.

VicDeo:
@PVince81 I hadn't got any feedback so I decided to do both core fix and a hotfix that will provide a seamless upgrade experience and potentially reduce number of support cases.
Surely it's up to management to decide whether hotfix is needed or not. :)

PVince81:
not wanting to merge this. People should update to 10.3 when it's out.

if needed, this PR can be sent as patch.

@pmaier1 @cdamken FYI

pmaier1:
Please make sure that there will be no issues in upgrading LDAP to 0.14.0, also on server versions below 10.3.

PVince81:
@VicDeo please adjust based on comments

maybe also add a PHP comment somewhere as a note that the command is not in info.xml as a workaround, for the future.

we'll need to keep in mind to revert this again for the next LDAP app version 0.15.0 or later

VicDeo:
@PVince81 updated, added fixme

PR #465, opened 2019-11-05
Fix only seeing one group per LDAP user.

Circuitsoft:
Without this change, each user gets only one LDAP group.


mmattel:
@jvillafanez @phil-davis adding so this PR can be reviewed

jvillafanez:
Latest master code works without this change.

Screenshot from 2020-02-18 14-30-38

Screenshot from 2020-02-18 14-31-00

There are no steps to reproduce nor setup to check, so I just checked the basic scenario we have for LDAP.
In addition, just setting a limit of 1000 groups isn't a good solution. What if there are 2000 groups? what if the a user is a member of 2000 groups?

Unless I'm missing something, I vote for closing.

mmattel:
Thanks for this analysis and feedback. Agree to close this PR.

phil-davis:
@Circuitsoft any feedback on the comments above?
Or can you give more information about reproducing the problem that you are fixing?

Circuitsoft:
Using Samba 4.9.5 as the LDAP server, it seems that an "unlimited" request stops after 1 group.

jvillafanez:
As far as I know, Samba uses an "Active Directory-like", so nested groups are expected there. Please, make sure you have that checkbox active in the configuration (advanced tab -> directory settings -> nested groups)

DeepDiver1975:
old -> dead -> close

PR #470, opened 2019-11-12
Detect a new DN by UUID attribute

VicDeo:
## Summary
I'm reusing an orphaned method `OCA\User_LDAP\User\Manager::dnExistsOnLDAP` to search the missing DN by UUID mapping
Please note that **UUID Attribute for Users** aka  `ldap_expert_uuid_user_attr`  should have really unique values.

## Steps
0. Prepare the following structure:
```
ou=parent,dc=owncloudqa,dc=com
  ou=unit1,ou=parent,dc=owncloudqa,dc=com
  ou=unit2,ou=parent,dc=owncloudqa,dc=com
```
1. Put users into  `ou=unit1,ou=parent,dc=owncloudqa,dc=com`
2. Setup user_ldap app:
add `ou=parent,dc=owncloudqa,dc=com` as a `ldap_base_users` 
In the expert tab add `UUID Attribute for Users` e.g. to be `uid`
3.  Execute `php occ user:sync -m remove "OCA\User_LDAP\User_Proxy"`
4. Login as any Ldap user, upload a new file and logout
5. **Move** this user from `ou=unit1,ou=parent,dc=owncloudqa,dc=com` to `ou=unit2,ou=parent,dc=owncloudqa,dc=com`
6.  Execute `php occ user:sync -m remove "OCA\User_LDAP\User_Proxy"`
7. Login as this user once again

### Expected
the user has a file uploaded on the step 4

### Actual 
the user has a skeleton only, firstrunwizard popup appears

## Issue
Fixes https://github.com/owncloud/enterprise/issues/3590

## There are some open questions:
- [x] Q. Should it always try to resole a missing DN by UUID or it is not desired in some cases? 
As it is possible to set any attribute  as UUID this attribute could be non-unique sometimes.
A. If duplication is detected we log it and skip this user.

- [x] Q. What should happen if the LDAP connection is lost while searching a new DN by UUID in the middle of user sync process?
A. We stop the sync

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1480,8 +1480,9 @@ class Access implements IUserTools {
 	 *
 	 * @param string $uuid
 	 * @return string
-	 * @throws \OutOfBoundsException
-	 * @throws \OC\ServerNotAvailableException
+	 * @throws \LengthException when more than one DN matches given UUID
+	 * @throws \OutOfBoundsException when there is no DN matching given UUID
+	 * @throws \OC\ServerNotAvailableException on any LDAP connection error
 	 */
 	public function getUserDnByUuid($uuid) {
 		$uuidOverride = $this->connection->ldapExpertUUIDUserAttr;
@@ -1515,9 +1516,13 @@ class Access implements IUserTools {
 
 		$filter = $uuidAttr . '=' . $uuid;
 		$result = $this->searchUsers($filter, ['dn'], 2);
-		if (isset($result[0]['dn']) && \count($result) === 1) {
-			// we put the count into account to make sure that this is
-			// really unique
+
+		// we put the count into account to make sure that this is really unique
+		if (\count($result) > 1) {
+			throw new \LengthException($uuidAttr . " is specified as UUID attribute but has a value '{$uuid}' for multiple entries");
+		}
+
+		if (isset($result[0]['dn'])) {
 			return $result[0]['dn'][0];
 		}
 
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -267,27 +267,19 @@ class Manager {
 	}
 
 	/**
-	 * checks whether a user is still available on LDAP
+	 * Try to resolve missing DN by UUID
 	 *
 	 * @param string $dn
 	 * @return bool
-	 * @throws \Exception
-	 * @throws \OC\ServerNotAvailableException
+	 *          - false if there is no UUID matching DN
+	 *          - true if there is at least one UUID matching DN
+	 *
+	 * @throws ServerNotAvailableException
 	 */
-	public function dnExistsOnLDAP($dn) {
-
-		//check if user really still exists by reading its entry
-		if (!\is_array($this->access->readAttribute($dn, '', $this->getConnection()->ldapUserFilter))) {
-			$lcr = $this->getConnection()->getConnectionResource();
-			if ($lcr === null) {
-				throw new \Exception('No LDAP Connection to server ' . $this->getConnection()->ldapHost);
-			}
-
-			try {
-				$uuid = $this->access->getUserMapper()->getUUIDByDN($dn);
-				if (!$uuid) {
-					return false;
-				}
+	public function resolveMissingDN($dn) {
+		try {
+			$uuid = $this->access->getUserMapper()->getUUIDByDN($dn);
+			if ($uuid) {
 				$newDn = $this->access->getUserDnByUuid($uuid);
 				//check if renamed user is still valid by reapplying the ldap filter
 				if (!\is_array($this->access->readAttribute($newDn, '', $this->getConnection()->ldapUserFilter))) {
@@ -295,14 +287,19 @@ class Manager {
 				}
 				$this->access->getUserMapper()->setDNbyUUID($newDn, $uuid);
 				return true;
-			} catch (\Exception $e) {
-				$this->logger->logException($e, ['app' => self::class]);
-				return false;
 			}
+		} catch (\LengthException $e) {
+			// UUID attribute matches multiple DNs - requires an action from LDAP administrator
+			$this->logger->warning($e->getMessage(), ['app' => 'user_ldap']);
+			// Keep user account until duplication is resolved
+			return true;
+		} catch (\OutOfBoundsException $e) {
+			// UUID attribute can't be detected this means that we need to drop the user on sync
+			$this->logger->info("DN '$dn' does not exist any more and can not be resolved using uuid attribute", ['app' => 'user_ldap']);
 		}
-
-		return true;
+		return false;
 	}
+
 	/**
 	 * returns an internal ownCloud name for the given LDAP DN, false on DN outside of search DN
 	 * @param UserEntry $userEntry user entry object backed by an ldap entry
--- a/tests/unit/User/ManagerTest.php
+++ b/tests/unit/User/ManagerTest.php
@@ -550,4 +550,74 @@ class ManagerTest extends \Test\TestCase {
 
 		self::assertSame('lastname', $this->manager->resolveUID($entry));
 	}
+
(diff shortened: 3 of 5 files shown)


VicDeo:
@cdamken Could you help me with open questions in https://github.com/owncloud/user_ldap/pull/470#issue-339901499

lib/User/Manager.php
```diff
@@ -267,42 +267,32 @@ public function getCachedEntry($uid) {
...
 				return true;
-			} catch (\Exception $e) {
-				$this->logger->logException($e, ['app' => self::class]);
-				return false;
 			}
+		} catch (ServerNotAvailableException $e) {
+			// A. pass? Acc will deleted then.
+			// B. fail? in the middle of the user sync?
+		} catch (\OutOfBoundsException $e) {
+			// UUID attribute can't be detected this means that we need to drop the user
```
jvillafanez: It seems that OutOfBoundsExceptions can also be thrown if the `getUserDnByUuid` finds 2 or more users instead of only one. We'd need to decide what to do in that case

lib/User/Manager.php
```diff
@@ -267,42 +267,32 @@ public function getCachedEntry($uid) {
...
 				}
 				$this->access->getUserMapper()->setDNbyUUID($newDn, $uuid);
 				return true;
-			} catch (\Exception $e) {
-				$this->logger->logException($e, ['app' => self::class]);
-				return false;
 			}
+		} catch (ServerNotAvailableException $e) {
+			// A. pass? Acc will deleted then.
+			// B. fail? in the middle of the user sync?
```
jvillafanez: I don't think we can handle that scenario here... What could happen if we let it bubble up? Any other place where we could handle that scenario?

tests/unit/User/ManagerTest.php
```diff
@@ -550,4 +550,49 @@ public function testResolveUID() {
...
+			->with($oldDn)
+			->willReturn($uuid);
+
+		$this->access->expects($this->any())
+			->method('getUserMapper')
+			->willReturn($mapper);
+		$this->access->expects($this->any())
+			->method('getUserDnByUuid')
+			->with($uuid)
+			->willReturn($newDn);
```
jvillafanez: Maybe we should be a bit more explicit, specially because this is a copy and paste of the test above for the most part (which is totally fine)
```
$this->access->expects($this->any())
			->method('readAttribute')
			->willReturn(null);
```

lib/User/Manager.php
VicDeo: ServerNotAvailableException is propagated to the Symfony application and user sync is aborted:
```
> php occ user:sync -m remove "OCA\User_LDAP\User_Proxy"
Analysing known accounts ...
   27 [--------------------------->]
In Manager.php line 278:
        
  Boom  
        

user:sync [-l|--list] [-u|--uid UID] [-s|--seenOnly] [-c|--showCount] [-m|--missing-account-action MISSING-ACCOUNT-ACTION] [-r|--re-enable] [--] [<backend-class>]
```
This is bad according to @cdamken as user sync is done in background and nobody will notice this.

tests/unit/User/ManagerTest.php
VicDeo: done

lib/User/Manager.php
VicDeo: If UUID is resolved into 2 different DNs it violates the requirement of uniqueness for UUID. 
As for me, we need to log a warning containing UUID and both DNs. Because this collision needs an action from LDAP administrator side.
In this case user should neither be deleted, nor remapped to any of  these new DNs, but old (missing) user DN should be left  "as is" until the collision is resolved.
@cdamken what do you think?

lib/User/Manager.php
jvillafanez: From my point of view, I think the admins should wrap the command in a way it suit their needs. They can check that the command has crashed if it ends with a non-0 status code (I guess we should be returning 1 now), so in this situation they can send a email using any command-line tool so they're notified in order to take any action.

I think the command must crash. There is nothing we can do to fix this situation. The only question that remains is if we should send somehow an email to the admin group, but this assumes at least one of the admins has set a valid email and that such admin can fix the problem (maybe the admin can't access to the host but just to ownCloud).
Note that the sync service will also increase the complexity. There is already a lot of logic there. Adding more logic implies a higher probability of bugs happening around something quite important as syncing the users.

lib/User/Manager.php
VicDeo: @jvillafanez I chaged the implementation to throw `LengthException` when more than one DN have been found.

lib/User/Manager.php
VicDeo: >we should be returning 1 now

exactly:
```
> echo $?
1
```
Updated to crash on `ServerNotAvailableException`

lib/Access.php
```diff
@@ -1515,9 +1515,13 @@ public function getUserDnByUuid($uuid) {
 
 		$filter = $uuidAttr . '=' . $uuid;
 		$result = $this->searchUsers($filter, ['dn'], 2);
-		if (isset($result[0]['dn']) && \count($result) === 1) {
-			// we put the count into account to make sure that this is
-			// really unique
+
+		// we put the count into account to make sure that this is really unique
+		if (\count($result) > 1) {
+			throw new \LengthException($uuidAttr . " is specified as UUID attribute but has a value '{$uuid}' for multiple entries");
```
jvillafanez: We need to update the PHPDocs and clarify when this exception is thrown. Note that I personally don't like to use predefined exceptions unless we throw them under the same assumptions they're expected to be thrown, which isn't the case, so the clarification is particularly important.

lib/User/Manager.php
jvillafanez: We must take any action here, even if it's just logging the error, otherwise the admin won't notice the problem and it could cause issues that won't be possible to detect.

lib/User/Manager.php
```diff
@@ -267,42 +267,36 @@ public function getCachedEntry($uid) {
...
 				return true;
-			} catch (\Exception $e) {
-				$this->logger->logException($e, ['app' => self::class]);
-				return false;
 			}
+		} catch (\LengthException $e) {
+			// UUID attribute matches multiple DNs - requires an action from LDAP administrator
+			$this->logger->warning($e->getMessage());
+			// Keep user account until duplication is resolved
+			return true;
```
jvillafanez: Just asking if there is any weird case we should be careful about.
We'd probably need to include PHPDocs explaining when this functions returns true or false.

lib/User/Manager.php
VicDeo: missing DN does not look like a problem - user can be deleted by LDAP admin so this exception is expected for a valid reason. Added a log entry with `info` severity

lib/Access.php
VicDeo: done, see above

lib/User/Manager.php
VicDeo: The explanation was added

lib/User/Manager.php
cdamken: The UUID is unique and the only case that could 2 different DN's appears is that one is not valid anymore and update the one who is new.

VicDeo: @cdamken "UUID is unique" is a requirement. But this requirement can be violated in the real life. Instead of just considering this situation as unbelievable I'd better add a piece of code to handle it explicitly. 

uuid attribute can be set to **any** ldap entry attribute  by the owncloud admin using `ldap_expert_uuid_user_attr`. Even to the attribute that have same values for **all** ldap users.

Dropping a user or random remapping is not a valid approach in this case. We should do nothing and log a warning about a duplication.

VicDeo:
no idea how a bunch of PHPDoc updates and 2 logger invocations could fail acceptence tests..

PR #473, opened 2019-11-26
Add drone setup for running tests with ldap and encryption enabled

dpakach:
### Add drone configuration for running core tests with LDAP and encryption app enabled.
Currently we run all the core tests with LDAP. In encryption app we run with encyption app enabled. But in more realistic live production environment, the system might have both LDAP and encryption at the same time. To test such scenarios this PR adds the drone configuration for running all the core tests with LDAP and encryption  enabled at the same time.

Related https://github.com/owncloud/qa-enterprise/issues/141#issuecomment-555869863

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -601,6 +601,470 @@ config = {
 				}
 			],
 		},
+		'core-api-acceptance-encryption-userkeys-nightly': {
+			'suites': {
+				'apiAuth': 'apiAuth-enc-UK',
+				'apiAuthOcs': 'apiAuthOcs-enc-UK',
+				'apiCapabilities': 'apiCapabilities-enc-UK',
+				'apiComments': 'apiComments-enc-UK',
+				'apiFavorites': 'apiFavorites-enc-UK',
+				'apiMain': 'apiMain-enc-UK',
+				'apiSharees': 'apiSharees-enc-UK',
+				'apiShareManagement': 'apiShareManagement-enc-UK',
+				'apiShareManagementBasic': 'apiShareMgmtBasic-enc-UK',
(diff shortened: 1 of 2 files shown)


.drone.starlark
```diff
@@ -601,6 +601,418 @@ config = {
...
+			],
+			'phpVersions': [
+				'7.1',
+			],
+			'runCoreTests': True,
+			'cron': 'nightly',
+			'extraApps': {
+				'encryption': ''
+			},
+			'extraSetup': [
```
phil-davis: IMO this will also need the `configure-app` `extraSetup` block like in `core-api-federation-encryption-userkeys-nightly` because it will not get added from the `extraSetup` in `defaults`

.drone.starlark
```diff
@@ -601,6 +601,418 @@ config = {
...
```
phil-davis: IMO this will also need the `configure-app` `extraSetup` block like in `core-api-federation-encryption-userkeys-nightly` because it will not get added from the `extraSetup` in `defaults`

.drone.starlark
```diff
@@ -601,6 +601,418 @@ config = {
...
```
phil-davis: IMO this will also need the `configure-app` `extraSetup` block like in `core-api-federation-encryption-userkeys-nightly` because it will not get added from the `extraSetup` in `defaults`

.drone.starlark
```diff
@@ -601,6 +601,418 @@ config = {
...
```
phil-davis: IMO this will also need the `configure-app` `extraSetup` block like in `core-api-federation-encryption-userkeys-nightly` because it will not get added from the `extraSetup` in `defaults`

Issue #475, opened 2019-11-29
phpunit tests failing on phpunit7


phil-davis:
https://drone.owncloud.com/owncloud/user_ldap/2005/5/4
```
phpdbg -qrr -d memory_limit=4096M -d zend.enable_gc=0 "/var/www/owncloud/server/apps/user_ldap/../../lib/composer/bin/phpunit" --configuration ./phpunit.xml --testsuite unit
PHPUnit 7.5.17 by Sebastian Bergmann and contributors.

Runtime:       PHPDBG 7.1.33-1+ubuntu18.04.1+deb.sury.org+1
Configuration: /var/www/owncloud/server/apps/user_ldap/phpunit.xml

.........................................E.....................  63 / 174 ( 36%)
............................................................... 126 / 174 ( 72%)
................................................                174 / 174 (100%)

Time: 1.27 seconds, Memory: 84.00 MB

There was 1 error:

1) OCA\User_LDAP\Controller\ConfigurationControllerTest::testCreateWithCopy
TypeError: Return value of PHPUnit\Framework\Constraint\Callback::matches() must be of the type boolean, object returned

/var/www/owncloud/server/apps/user_ldap/lib/Configuration.php:437
/var/www/owncloud/server/apps/user_ldap/lib/Configuration.php:307
/var/www/owncloud/server/apps/user_ldap/lib/Controller/ConfigurationController.php:104
/var/www/owncloud/server/apps/user_ldap/tests/unit/Controller/ConfigurationControllerTest.php:140

ERRORS!
Tests: 174, Assertions: 424, Errors: 1.
```

PR #476, opened 2019-11-29
Fix testCreateWithCopy PHP unit test

phil-davis:
Issue #475

### DIFF
--- a/tests/unit/Controller/ConfigurationControllerTest.php
+++ b/tests/unit/Controller/ConfigurationControllerTest.php
@@ -128,7 +128,7 @@ class ConfigurationControllerTest extends TestCase {
 							break;
 						default: $expectedValue = null;
 					};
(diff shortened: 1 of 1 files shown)


tests/unit/Controller/ConfigurationControllerTest.php
```diff
@@ -128,7 +128,7 @@ public function testCreateWithCopy() {
 							break;
 						default: $expectedValue = null;
 					};
-					return $this->stringStartsWith('tgt');
```
phil-davis: `stringStartsWith` returns an object that is a "constraint" object within `phpunit` `Assert`. 
Actually here we are supposed to return a boolean that indicates if `$key` starts with `tgt`

phil-davis:
codecov is slightly failing ` 35.95% (-0.02%) compared to 17c2f26`
Somehow it is too stupid - IMO we can override that.

PR #490, opened 2019-12-19
Allow plus in LDAP usernames

phil-davis:
core PR https://github.com/owncloud/core/pull/36613 proposes allowing `+` in usernames.

If that PR is accepted, then we should/could also allow `+` in usernames coming from LDAP.

This PR implements that, and adds unit tests to check the LDAP `sanitizeUsername()` method. (There were not any tests for that yet)

- [x] set minimum core version to 10.4 (we only want to allow sync of LDAP usernames containing `+` if core also supports usernames containing `+`)

IMO this should only be merged and released after core 10.4 has been released.


phil-davis:
@pako81 this would also be needed if we start allowing `+` in usernames/UIDs

pako81:
@phil-davis makes sense 👍

phil-davis:
CI here is "stuck" - with core min version 10.4 we cannot have test pipelines against latest.
But there is no rush - we can run CI again when core 10.4 is latest.

mmattel:
@phil-davis has there anything changed that allows a progress 😅

phil-davis:
I rebased to catch up with current user_ldap.

There is no point releasing this anyway, until core 10.4 is out. So we can easily merge this the day that 10.4 community release happens and release user_ldap soon after.

phil-davis:
@micbar @HanaGemela what do you want to do about this PR?

I could adjust the CI here so it does not test against `latest` and bumps core min-version to 10.4. Then CI will pass and we could merge it and prepare the app release before official core 10.4 release - like we are doing for various other apps.

micbar:
@phil-davis Does it really have a strickt dependency? what would happen, if I use the current version with 10.4.0?

phil-davis:
If your LDAP does not actually have any usernames containing `+` then yes, it should still work on a core 10.2 or 10.3 system.

Do you want to try and keep the new version of user_ldap installable on 10.2 and 10.3?

phil-davis:
Note: if we wait until core 10.4.0 becomes the "latest" community tarball, then I can remove the last commit from this PR.

phil-davis:
> Note: if we wait until core 10.4.0 becomes the "latest" community tarball, then I can remove the last commit from this PR.

10.4.0 is now "latest". I reverted the 3rd commit. Let's see what happens to CI.

mmattel:
Green, green :smile:

phil-davis:
@micbar please get a review of this.

PR #495, opened 2020-01-10
Migrate ldap connection and configurations from user_ldap tests to core tests

kiranparajuli589:
## Description
It was a bit annoying to interact with zend-ldap from user_ldap tests.

Setups and configuration for ldap-connection has been migrated to core tests!

### DIFF
--- a/tests/acceptance/features/apiUserLDAPConnection/groupFilter.feature
+++ b/tests/acceptance/features/apiUserLDAPConnection/groupFilter.feature
@@ -4,7 +4,18 @@ Feature: filter groups
   I want to be able to filter LDAP groups
   So that only groups meeting specific criteria are available in ownCloud
 
-    Scenario: single group filter
+  Background:
+    Given these groups have been created:
+      | groupname    |
+      | grp1         |
+      | grp2         |
+      | group1       |
+      | group2       |
+      | ShareeGroup  |
+      | ShareeGroup2 |
+
+
+  Scenario: single group filter
     When the administrator sets these settings of LDAP config "LDAPTestId" using the occ command
       | key             | value                                        |
       | ldapGroupFilter | (&(\|(objectclass=posixGroup))(\|(cn=grp2))) |
@@ -15,7 +26,7 @@ Feature: filter groups
       | admin   |
       | grp2    |
 
-    Scenario: filter with asterisk 
+  Scenario: filter with asterisk
     When the administrator sets these settings of LDAP config "LDAPTestId" using the occ command
       | key             | value                                           |
       | ldapGroupFilter | (&(\|(objectclass=posixGroup))(\|(cn=Sharee*))) |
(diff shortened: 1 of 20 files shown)


phil-davis:
this will need a rebase - there were CI fixes merged in another PR this morning.

tests/acceptance/features/bootstrap/UserLdapGeneralContext.php
```diff
@@ -318,6 +320,118 @@ public function getLdapPort() {
...
+	 *
+	 * @param TableNode $table
+	 *
+	 * @return void
+	 * @throws LdapException
+	 * @throws Exception
+	 */
+	public function followingUsersHaveBeenCreatedWithoutSkeletonFilesUsingLDAPServer($table) {
+		$table = $table->getColumnsHash();
+		$setDefaultAttributes = '';
```
dpakach:
```suggestion
		$setDefaultAttributes = false;
```
I think it would be better to use this variable as a boolean.

dpakach:
Overall looks good. But one concern I have is that when we load users directly from the ldif file then they are also added to certain groups automatically. When doing this we lose that I suppose. How can we handle that?

phil-davis:
Yes, we are going to have to also use methods for creating/deleting groups in LDAP.
And we have to sort out where these LDAP user/group create/delete methods go (e.g. in core?) because they need to be called from the core steps that create/delete users.

That might work OK to call from core into the user_ldap context file, because the user_ldap methods will only be called when running user_ldap tests. Those methods will not exist when running core tests (or other app tests) but things will not crash because the method never gets called?

.drone.starlark
```diff
@@ -2218,7 +2218,7 @@ def installCore(version, db, useBundledApp):
 		'image': 'owncloudci/core',
 		'pull': 'always',
 		'settings': {
-			'version': version,
+			'git_reference': 'ldap-auto-complete',
```
kiranparajuli589: only to test drone locally..

tests/acceptance/features/apiUserLDAPConnection/groupFilter.feature
```diff
@@ -36,25 +39,24 @@ Feature: filter groups
...
-      """
-      When the administrator sets these settings of LDAP config "LDAPTestId" using the occ command
-        | key             | value                                                     |
-        | ldapGroupFilter | (&(\|(objectclass=posixGroup))(\|(cn=grp1))) |
-      And the administrator gets the groups in JSON format using the occ command
-      Then the command should have been successful
-      And the groups returned by the occ command should be
-        | group  |
-        | admin  |
-        | grp1   |
```
kiranparajuli589: just wrong indentation fix!

tests/acceptance/features/apiUserLDAPConnection/groupFilter.feature
```diff
@@ -15,7 +15,7 @@ Feature: filter groups
       | admin   |
       | grp2    |
 
-    Scenario: filter with asterisk
```
kiranparajuli589: make-test-php-style-fix

PR #510, opened 2020-02-09
Update starlark code 20200208

phil-davis:
Issue https://github.com/owncloud/QA/issues/640


phil-davis:
I had some strange problem when I tried to use `.drone.star`
https://drone.owncloud.com/owncloud/user_ldap/2236
```
yaml: did not find expected hexdecimal number
```

First let's get the starlark code changes in, then I can investigate switching to `.drone.star`

There are many pipelines generated in this repo, so it might be overflowing something in the starlark on the drone server...

Issue #515, opened 2020-02-12
user:sync -u is buggy


voroyam:
If you try to sync a user, that does not exist, the message is the same as if he existed and does not exist anymore. ownCloud asks what to do and you can delete or disable.

How can you remove something that was not there in the first place???

Who tested this new command flag?

phil-davis:
Also see issue #511 - `user:sync` of a user that is actually a local user (never came from LDAP in the first place).

We are adding some/more/better acceptance test coverage of `user:sync` https://github.com/owncloud/core/issues/36597

We will add test scenarios like this, to demonstrate "bad behaviour"

voroyam:
I would like to know how this was tested before release?

Or do we just release stuff without testing it?

phil-davis:
Looks `-u` was introduced in PR https://github.com/owncloud/core/pull/30485 (backport to stable10 https://github.com/owncloud/core/pull/31025 ) in April 2018. So it is not so "new" - been there sine 10.0.8 by the look of it.

Back then we did not have automated test coverage for `occ` commands. I can't say exactly what scenarios were manually tested at the time (or if it worked then and now has a problem/regression in behaviour). Over time we are expanding automated test coverage. We have recently been adding scenarios for the `user:sync` command. The automated test scenarios will ensure that regressions do not easily happen.

voroyam:
I like what you wrote. I would just write "no idea" but that's unprofessional :)

phil-davis:
PR #517 has added a test scenario that demonstrates this "feature/bug".

When someone adjusts the behavior, they will need to adjust the test scenario also and remove the issue tag.

PR #517, opened 2020-02-17
[Tests-Only] Added acceptance test for user:sync with non-existing user

kiranparajuli589:
## Description
Added acceptance test for occ command `user:sync` for a non-existing user

## Related Issue
- Reproduces https://github.com/owncloud/user_ldap/issues/515
- Part of https://github.com/owncloud/core/issues/36597

### DIFF
--- a/tests/acceptance/features/cliProvisioning/userSync.feature
+++ b/tests/acceptance/features/cliProvisioning/userSync.feature
@@ -85,7 +85,6 @@ Feature: sync user using occ command
 
   @issue-515
   Scenario: sync a user that does not exist
-    Given user "regularuser" should not exist
     When LDAP user "regularuser" is resynced
     Then the command should have been successful
     And the command output should be:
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/cliProvisioning/userSync.feature
```diff
@@ -82,3 +82,15 @@ Feature: sync user using occ command
     And user "regularuser" should exist
     And the display name of user "regularuser" should be "Test User"
     And the display name of user "regular" should be "Regular User"
+
+  @issue-515
+  Scenario: sync a user that does not exist
+    Given user "regularuser" should not exist
```
HariBhandari07: I think the given step is not required

kiranparajuli589: Then how can we say the user is not created before or it doesn't exist??

HariBhandari07: I think you can also show the command output for already existing user

HariBhandari07: @kiranparajuli589 user0 and user1 are already created in the background, so you can also add steps to show syncing of one of them while regular user is not created we may already have some scenarios to check the existence of not created user. So checking for the existence of non-existent user is redundant.

kiranparajuli589: @HariBhandari07 I've removed the `given` step!
Please review...

PR #523, opened 2020-02-26
Update dependencies

phil-davis:
```
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 0 updates, 1 removal
  - Removing zendframework/zend-ldap (2.10.1)
Writing lock file
Generating autoload files
```

It seems a bit odd that composer removes `zendframework/zend-ldap` but does not replace it with `laminas/laminas-ldap`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ clean: clean-composer-deps clean-dist clean-build
 
 .PHONY: clean-composer-deps
 clean-composer-deps:
+	rm -Rf vendor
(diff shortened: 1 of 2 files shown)


phil-davis:
Hmmm - PR #495 removed `zendframework/zend-ldap` from `composer.json`
A new `composer.lock` was never generated, so it did not immediately break anuthing.
I will make a PR to put it back.

PR #533, opened 2020-03-16
[Tests-Only] merge api tests for LDAP into a single suite

dpakach:
Merge API test suites for ldap into single suites

### context
Most of these suites have very few test scenarios but while running tests in drone, each suite requires its its own setup processes. So in order to minimize the run time in CI, having less number of suites is better.


phil-davis:
Needs more sorting out context files: https://drone.owncloud.com/owncloud/user_ldap/2374/31/11

PR #539, opened 2020-04-08
Set core min_version to 10.3 with PHP min of 7.1

phil-davis:
We could allow user_ldap with core 10.3.2 still, if we want. It should work fine.
It is only #490 "Allow plus in LDAP usernames" that will cause an issue if someone has user_ldap with core 10.3.2 and then does create an LDAP username that contains a `+` sign.

Note: we should do the `composer.json` change at some time anyway. If this PR  does not happen, then I will do that to `master` for some later release.


davitol:
Final decision was to keep 10.4 min version for the app. Thanks anyway @phil-davis and sorry for the noise.

Issue #250, opened 2018-06-14
load users and groups fully into acceptance tests


phil-davis:
Currently the users and groups used by the user_ldap acceptance tests are stored in ``ldap-users.ldif`` and that is loaded as the starting point for a test scenario.

The test scenario underlying code needs to be aware of what users and groups "magically exist" so that it can correctly know what to expect when sharing, examining autocomplete matches etc.

In PR #249 we explicitly pasted in the list of users and group in the test scenario setup steps.

Rather than do that, it will be cleaner if the code the loads in ``ldap-users.ldif`` also "tells" the test code about the users and groups it has "created".

phil-davis:
We need to discuss and sort out the dependency between the core test scenarios, that mention various users, groups and group membership, and the "fixed" set of users, groups and group membership that exists in user_ldap CI.

There is getting to be more non-obvious "magic use of particular users and groups" in the core test scenarios that is fragile and easily breaks user_ldap testing.

phil-davis:
More of this annoying stuff happened. See "fix" in https://github.com/owncloud/core/pull/35746

It would be great to find a better way to organize this.

phil-davis:
And also https://github.com/owncloud/core/issues/36480 https://github.com/owncloud/core/pull/36481
It would be nice to have a way that users can be (optionally) existing/created at the start of a scenario without them being initialized (i.e. so that they do not already have the skeleton files existing)

phil-davis:
I think we fixed this in the recent refactoring. PR #543 removes a last reference,  then we can close this issue?

PR #543, opened 2020-04-13
Remove ldapInitialUserFilePath reference to ldap-users.ldif

phil-davis:
Part of issue #250

`ldapInitialUserFilePath` is not used anywhere anymore. Remove it to avoid confusion.

Note: `ldap-users.ldif` file is also not used any more. I have left it there for now as an example `ldif` file that developers-testers can copy-paste from. But maybe it should be deleted, edited, renamed, or? To avoid confusion in the future for people who might think it is actually used in the tests.

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -11,7 +11,6 @@ default:
           ldapAdminPassword: admin
           ldapUsersOU: TestUsers
           ldapGroupsOU: TestGroups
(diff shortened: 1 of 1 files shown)


phil-davis:
My mistake. `ldapInitialUserFilePath` points to the core `ldap-users.ldif` which just has the `TestUsers`  and `TestGroups` empty Organizational Units. These are created empty at the start of tests, which is "a good thing".

PR #559, opened 2020-05-25
[Tests-Only] Removed unused variable from exception message

kiranparajuli589:
In PR #517, there was a mistake in refactoring. `getStdErrFromOccCommand` is used instead!


phil-davis:
Fails coding standard.

And then CI will fail anyway because of https://github.com/owncloud/core/issues/37428 - so we will have to fix it all together.

phil-davis:
This fix is included in #560

PR #557, opened 2020-05-20
Use plugins/codecov:latest

phil-davis:
From codecov support: "Can you please update to the latest version of the uploader you are using and try again please?"

PR #560, opened 2020-05-25
[Tests-Only] Added ldapAttributesForUserSearch in setConfig.sh

kiranparajuli589:
Fixes https://github.com/owncloud/core/issues/37428

Added `ldapAttributesForUserSearch` as `uid` inside `setConfig.sh`


phil-davis:
Why does https://github.com/owncloud-docker/compose-playground/blob/master/compose/ldap/ldap.sh have a list of settings that have underscores, like `ldap_base_groups`
But https://github.com/owncloud/user_ldap/blob/master/tests/acceptance/setConfig.sh has `ldapBaseGroups` ?

It would be nice to understand this!

kiranparajuli589:
> Why does https://github.com/owncloud-docker/compose-playground/blob/master/compose/ldap/ldap.sh have a list of settings that have underscores, like `ldap_base_groups`
> But https://github.com/owncloud/user_ldap/blob/master/tests/acceptance/setConfig.sh has `ldapBaseGroups` ?
> 
> It would be nice to understand this!

Both naming convention is supported for occ commands. We can use either one!

phil-davis:
I pushed the changes from PRs #559 and #557 - we might as well get all that merged in a single run of CI.

I also added a commit to "Fix expected output when syncing a user that does not exist"

Hopefully we get green CI!

kiranparajuli589:
@phil-davis there was output-message change in https://github.com/owncloud/core/pull/37398/files#diff-0ada4aa7ff5e95845c939f25bf35448dL253-R277
due to which 1 acceptance test was failing.
` Command output message is now also updated in acceptance tests`

phil-davis:
@kiranparajuli589 What changed when you pushed again the last commit?
I do not see what is different from the way it was.

mrow4a:
@phil-davis I think he refers to https://github.com/owncloud/user_ldap/pull/560/files#diff-bc1f2fd46d03b1424b0575c333d65291L92-R94

phil-davis:
yes, I know. I had pushed that a while ago. Maybe I made a small mistake - I just do not see how the newly-pushed commit is any different to what I originally pushed. Anyway, it does not matter - if it passes then it is good!

phil-davis:
https://drone.owncloud.com/owncloud/user_ldap/2536/48/11
Passes against core master.

https://drone.owncloud.com/owncloud/user_ldap/2536/53/11
But of course now fails against core "latest" which is 10.4.1.
I will add a commit to skip that scenario on old oC  versions.

PR #571, opened 2020-06-03
[Tests-Only][DoNotMerge] Refactor tests

jasson99:


jasson99:
This is already fixed in pr: #573

phil-davis:
I didn't realise that this PR was waiting. We could have just rebased it this morning. Sorry about that.

Anyway, it seems that user_ldap is now passing :rocket:

PR #581, opened 2020-06-22
Add warning for disabling email login regarding strict login check

mrow4a:
Adds a warning that this config is not effective if strict checking is not enabled. Requires  https://github.com/owncloud/core/pull/37569

image

### DIFF
--- a/templates/part.wizard-loginfilter.php
+++ b/templates/part.wizard-loginfilter.php
@@ -18,7 +18,7 @@
 			</label>
 
 			<input type="checkbox" id="ldap_loginfilter_email"
-				   title="<?php p($l->t('Allows login against an email attribute. Mail and mailPrimaryAddress will be allowed.'));?>"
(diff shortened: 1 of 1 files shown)


mmattel:
@pako81 @karakayasemi @HanaGemela please rebase and merge.

karakayasemi:
@phil-davis @mrow4a I just rebased the PR and CI failed. Also, there are several CI fails in nightly jobs. Looks like, one of the recent changes in core or this app, broke tests.

phil-davis:
> @phil-davis @mrow4a @micbar I just rebased the PR and CI failed. Also, there are several CI fails in nightly jobs. Looks like, one of the recent changes in core or this app, broke tests. Do you have any idea for the reason?

You are correct - it got broken with some core acceptance test code changes related to trying to cope with various combinations of LDAP stuff for OCIS testing. When running here in `user_ldap` it looks like the step:
```
And user "Alice" has been created with default attributes in the database user backend
```
now actually ends up creating `Alice` in LDAP. And then the display name cannot be changed "locally" in oC10.

The problem was not noticed because there is also some problem with pulling `osixia/openldap` and a lot of CI was failing in the setup steps, even before starting to run any tests. See #590 

I will sort out the problem in core...

Issue #583, opened 2020-07-02
sort out accceptance tests that are now failing in user_ldap


phil-davis:
PR https://github.com/owncloud/user_ldap/pull/582 and core PR https://github.com/owncloud/core/pull/37616 adjusted the tags used for selecting the core acceptance tests to be run from user_ldap.

The `TestAlsoOnExternalUserBackend` tag was removed. Probably there are more scenarios that are now being run in the nightly CI, and many of those should have `skipOnLDAP`.

Sort it out so CI passes.

phil-davis:
for example, last night CI https://drone.owncloud.com/owncloud/user_ldap/2633/223/14 (quite a few pipelines have fails)

phil-davis:
This should be fixed. If it fails again tonight we can easily open the issue again.

PR #585, opened 2020-07-03
[Tests-Only] Run only smokeTest for core-webui-acceptance-encryption-userkeys-nightly

phil-davis:
Part of issue #583 

The core webUI acceptance tests are run in the nightly job. With master-key encryption we run just the `smokeTest` scenarios. But accidentally we did not limit the scenarios for user-key encryption.

Make this consistent .

This will stop "extra" scenarios from being run, including various scenarios that are not expected to pass on LDAP anyway.


phil-davis:
This was already in PR #584 - I missed that.

PR #596, opened 2020-10-15
add integration tests for LDAP Access with many ou and testing baseDn configurations

mrow4a:
This PR adds integration tests for ldap server with many OU in base DN, and querying users across many bases.

### DIFF
--- a/tests/integration/Lib/IntegrationTestPagingMultipleBase.php
+++ b/tests/integration/Lib/IntegrationTestPagingMultipleBase.php
@@ -28,7 +28,7 @@ use OCA\User_LDAP\User_LDAP;
 
 require_once __DIR__ . '/../AbstractIntegrationTest.php';
 
-class IntegrationTestPaging extends AbstractIntegrationTest {
+class IntegrationTestPagingMultipleBase extends AbstractIntegrationTest {
 	/** @var  UserMapping */
 	protected $mapping;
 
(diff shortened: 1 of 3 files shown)


tests/integration/Lib/IntegrationTestPagingMultipleBase.php
```diff
@@ -0,0 +1,82 @@
...
+	 * @return bool
+	 */
+	protected function case1() {
+		$filter = 'objectclass=inetorgperson';
+		$attributes = ['cn', 'dn'];
+
+		// having multiple bases there is currently an issue that:
+		// - having searchUsers with limit 2 and offset 2, expected resultset is 2
+		// - while having multiple base-dns specified, LDAP currently will return 4 results (while expected is 2 from range query)
+		$result = $this->access->searchUsers($filter, $attributes, 2, 0);
```
mrow4a, line 62: @jvillafanez @VicDeo @cdamken I might need help here. When having multiple base DNs e.g. `$bdnUsers = 'ou=users,' . $bdn . ';' .'ou=admins,' . $bdn;` , we now return N times results then expected by limit/offset. Is this expected behaviour? Or this means we do not (ownCloud) support specifying multiple base dns per single LDAP server configuration  ?
95899960-94a21000-0d91-11eb-9cd0-4b92ced79aca

tests/integration/Lib/IntegrationTestPagingMultipleBase.php
```diff
@@ -0,0 +1,82 @@
...
+		echo(\count($result) . PHP_EOL);
+		if (\count($result) !== 4) {
```
mrow4a, line 64: we return 4 results for limit=2 offset=0 !

tests/integration/Lib/IntegrationTestPagingMultipleBase.php
jvillafanez: Either known bug or not supported. Anyway, fixing it would need to rewrite a good amount of LDAP code, and I'm not sure how many people would require to use multiple base dn.
I'd vote for not supported at this point.

mrow4a: @micbar option to consider -> add new commit that displays error in UI informing that multi-base DN per single server is currently not supported.

mrow4a:
moved to https://github.com/owncloud/user_ldap/pull/307

PR #597, opened 2020-10-22
Facelift

hurradieweltgehtunter:
Hi,
while working on an ldap implementation in a custom theme I update this app to have a more modern look.
Note: Those are only design changes. No functional changes were made.

# General look
- General look is now more unified
- Advanced/Expert tabs have a warning icon to emphasize that those are critical settings
- Connection test icon was moved to the tab bar to have it visible at all time (before users had to scroll down to check)
- Results of config tests are displayed always besides test buttons (before results were shown besides button or as OC Notification sometimes)
- Repsonsiveness was added (elements resize based on media queries)
- Correct use of HTML elements

before:
Bildschirmfoto 2020-10-22 um 12 59 52

after:
Bildschirmfoto 2020-10-22 um 15 21 50

## Fixed top bar
- Tab bar is a fixed element, tabs, test indicator icon, help are always visible now (see gif)
before:
scroll_before

after:
scroll

# Input labels & Hints
Placeholder values were the names of the fields. If a user entered a value the name (=placeholder) dissappeared. All input fields have labels now so users know anytime which input is expected.
Also helping tips were hidden in mouseover tooltips. Tips are now displayed below input fields to have them visible at any time.

before:
Bildschirmfoto 2020-10-22 um 13 00 58

after:
Bildschirmfoto 2020-10-22 um 13 01 53

# Advanced settings
Advanced settings were displayed as an accordion which hides some settings and the user is forced to open them manually.
Accordion was removed to have all settings visible at the same time.

before:
Bildschirmfoto 2020-10-22 um 13 02 52

after:
Bildschirmfoto 2020-10-22 um 13 02 35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+.
 
 ## [0.15.2] - 2020-06-16
 
--- a/css/settings.css
+++ b/css/settings.css
@@ -47,10 +47,6 @@
 	height: auto;
 }
 
-#ldap .tablerow button.ui-multiselect {
-
-}
-
 #ldap .tablerow .inline > div {
 	display: inline-block;
 }
@@ -327,6 +323,7 @@ select[multiple=multiple] + button {
 @media all and (min-width: 768px) {
 	#ldap .ui-tabs .ui-tabs-nav li.stateIndicator {
 		display: block;
+	}
 }
 
 .ui-tabs .ui-tabs-panel {
@@ -357,8 +354,6 @@ select[multiple=multiple] + button {
     padding-right: 5px;
 	margin-top: 1em;
 }
-
-
 
 #ldap p {
 	margin-bottom: 1em;
--- a/css/settings.css
+++ b/css/settings.css
@@ -1,24 +1,88 @@
+#ldap section {
+	margin-bottom: 2rem;
+	width: 100% !important;
+}
+
+@media all and (min-width: 768px) {
+	#ldap section {
+		width: 60% !important;
+	}
+}
+
+#ldap section h3 {
+	border-bottom: 1px solid #efefef;
+	margin-left: -41px;
+	padding-left: 41px;
+	padding-bottom: 5px;
+}
+
 .table {
 	display: table;
 	width: 85%;
 }
 
 .tablerow {
-	display: table-row;
-	white-space: nowrap;
 	text-align: left;
-}
-
-.tablerow input, .tablerow textarea {
-	width: 100% !important;
-}
-
-.tablerow textarea {
-	height: 15px;
+	margin-bottom: 1rem;
 }
 
 #ldap .tablerow label {
-	margin-left: 3px;
+	margin-left: 0;
+	display: inline-block;
+}
+
+#ldap .tablerow input, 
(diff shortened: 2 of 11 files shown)


mrow4a (approved):
from server-side perspective, looks good.

mrow4a:
@micbar might need QA testdrive and see for all buttons is values are correctly set in DB

voroyam:
> Advanced/Expert tabs have a warning icon to emphasise that those are critical settings

From my pov it looks like something is wrong in those tabs now. Those symbols are generally used for warnings. Like "warning something is burning over here"

----

I don't see the arrows for add and remove listed in the change log. It's confusing now.

before it was 

on the left are groups that are available - arrows - groups that are selected for sync
You can click on left or right now move one group from the one list to another. Makes sense.

now

image

I think `> <` looks like a smiley standing alone in the line like that. It's not clear what it does.

I would be more than happy to give you some input on the LDAP app, or any support person, we usually spend a lot of time working with this app, so maybe we have some input :) Just a suggestion.

hurradieweltgehtunter:
The idea was not to change too much and to not touch the config form because it would require a change in JS which would be very much effort. I agree, the `> <` buttons look strange but my first guess is that this wasn't changed by this PR.
Any effort beyond these small changes was declined because it would lead in a complete redesign of the wizard.

cdamken:
I think less work just to do something is not the right approach. 

I don't think the redesign was declined, it was already one that Felix Heidecke did, what happened with it?

IMHO, this one looks too minimalistic, Probably @tbsbdr could help with a better design and I could help with the use.

For example:
There was a reason why the `advanced` and the `expert` tabs were separated to the right.
The `expert` tab should be entered once and avoid changing it again. with a testing output before is fixed with a confirmation.

Probably some stuff in the `advanced` tab should be redefined and reorganize. for example, the storage should be in the expert tab in my opinion.

hurradieweltgehtunter:
What I was told back then, Felix did a complete rewrite which resulted in an unusable app because the settings and combination of settings were too complex.
My approach was simply improve the look by working on CSS - nothing else. Even moving settings from one tab to another would require a deeper look inside the app and would require LDAP knowledge. 
When I did those - at least intended - improvments noone was able to provide this knowledge because - again - too much effort. 
This effort was declined by product management.

Moving the two tabs to the left was done because they were outside the viewport on smaller screen sizes. The change made it - at least somewhat, given the HTML structure - responsive.
The red trianlges were included to indicate sensitive settings. I agree, it can be interpreted as an error. IMHO it should be pointed out somehow that those are sensitive. Using text blocks like "Careful, sensitive settings" would require a transifex change and a lot of translations. That's why I decided against it, also to keep the effort as low as possible.

If y'all have a **good** and suitable solution I'll be happy to implement it. The current state was my best try considering all given facts.

cdamken:
Well, if PM (@hodyroff and @pmaier1 ) decided that way I'm just have to accept it.

Any way, then we have to fix the `> <` smilie that is really bad!

tbsbdr:
@hurradieweltgehtunter thanks for the facelift! I guess the >< needs some rework. Perhaps a wording like "+ add" "-remove" could help? but i'm not familiar with the usecase/whole picture... Could you make a proposal? @voroyam @cdamken you should be now in the loop

voroyam:
My proposal would be to revert it back how it was before: 
2 columns left and right, in the middle the buttons. Their placement requires no explanation.

If the motto of the day is to invest to work - then going back is least amount of work I could imagine. anything else requires suggestions and revisions and approvals and concepts.

PR #599, opened 2020-10-30
Fixed read LDAP attribute value 0 returned as null

mojontwins:
I noticed that when sync'ing LDAP users with a quota of "0", they were stored with a quota of NULL. The reason is that there's a soft comparison here:

```
	protected function walkBackends($uid, $method, $parameters) {
		$cacheKey = $this->getUserCacheKey($uid);
		foreach ($this->backends as $configPrefix => $backend) {
			$instance = $backend;
			if (!\method_exists($instance, $method)
				&& \method_exists($this->getAccess($configPrefix), $method)) {
				$instance = $this->getAccess($configPrefix);
			}
			if ($result = \call_user_func_array([$instance, $method], $parameters)) {
				$this->writeToCache($cacheKey, $configPrefix);
				return $result;
			}
		}
		return false;
	}
```

So when call_user_func_array returns "0" it is considered "false", the value is not cached nor returned, and the method returns "false".

Just adding !== false solves the problem. Now my users with quota "0" are synchronized correctly and they show "0" in the oc_accounts table rather than NULL.

### DIFF
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -81,7 +81,7 @@ class User_Proxy extends Proxy implements
 				&& \method_exists($this->getAccess($configPrefix), $method)) {
 				$instance = $this->getAccess($configPrefix);
 			}
-			if ($result = \call_user_func_array([$instance, $method], $parameters)) {
+			if (($result = \call_user_func_array([$instance, $method], $parameters)) !== false) {
 				$this->writeToCache($cacheKey, $configPrefix);
 				return $result;
 			}


mmattel:
any update on this?
adding @butonic as reviewer

mmattel:
@phil-davis sorry to bother you, can you take a look why codecov is not starting?
Would like to make that green to ease reviews :smile:

phil-davis:
Rebased to get fresh CI.
codecov has been replaced by SonarCloud. Those codecov statuses should no longer be "required". I am waiting for someone to change the repo settings. For now, I will "fake" the codecov results.

jnweiger:
@jvillafanez Should this also solve https://github.com/owncloud/user_ldap/pull/465 ?

jvillafanez:
This is for user operations. https://github.com/owncloud/user_ldap/pull/465 is for groups operations. It seems unrelated.

In addition, as commented in https://github.com/owncloud/user_ldap/pull/465, it's unclear that PR will fix the issue.

PR #600, opened 2020-11-02
Adjust PHP dependencies for composer 2.0

phil-davis:
Issue https://github.com/owncloud/core/issues/38067

See issue for details - `behat` acceptance tests need to use PHP  7.4 for the dependencies to sort themselves out with composer 2.0

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -1724,7 +1724,7 @@ def acceptance():
 									},
 									'steps':
 										installCore(server, db, params['useBundledApp']) +
-										installTestrunner(phpVersion, params['useBundledApp']) +
+										installTestrunner('7.4', params['useBundledApp']) +
 										(installFederated(server, phpVersion, params['logLevel'], db, federationDbSuffix) + owncloudLog('federated') if params['federatedServerNeeded'] else []) +
 										installApp(phpVersion) +
 										installExtraApps(phpVersion, params['extraApps']) +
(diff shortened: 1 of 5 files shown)


.drone.starlark
```diff
@@ -1906,6 +1906,7 @@ def ldapService(ldapNeeded):
 				'LDAP_ORGANISATION': 'owncloud',
 				'LDAP_ADMIN_PASSWORD': 'admin',
 				'LDAP_TLS_VERIFY_CLIENT': 'never',
+				'HOSTNAME': 'ldap',
```
phil-davis, line 1909: Note: other core app drone has `HOSTNAME` defined here. So this is being made consistent with other CI.

composer.json
```diff
@@ -1,20 +1,20 @@
 {
     "config": {
         "platform": {
-            "php": "7.0.8"
+            "php": "7.1"
```
phil-davis, line 4: The min core version for `user_ldap` is already set to 10.4, and that requires at least PHP 7.1. So this change does not effect which core ownCloud10 that user_ldap will install on.

micbar (approved):
Why does it have a `.drone.yml` checked in?

Rest looks fine.

phil-davis:
> Why does it have a .drone.yml checked in?

There is trouble with the generated file size if we let drone convert `.drone.star` on the server. The extension that does that has an artificial limit of 1,000,000 bytes for the internally-generated YML. We haven't found a way around that yet. Brad at drone would not accept changing that to 2,000,000.

.drone.starlark
phil-davis: didn't work - reverted.

Issue #616, opened 2021-02-01
Nightly CI failing some user-key encryption tests


phil-davis:
Encryption PR https://github.com/owncloud/encryption/pull/25 was merged a week or so ago. That requires matching code changes in core `master` that are not available in 10.6.0 release. Some nightly tests in `user_ldap` have the encryption app also enabled, and run with `latest` core (which is currently core 10.6.0). Those fail, and we understand why.

- [x] Adjust the nightly CI so that pipelines with encryption uses current core master for now. #618 
- [x] Make an issue to change the testing back after 10.7.0 is released (so that we do not forget) #620

phil-davis:
Actually that PR #617 probably does not fix all of this. Let's see the results of the nightly CI tomorrow.

PR #619, opened 2021-02-05
[tests-only] Test encryption pipelines

phil-davis:
To see if issue #616 is fixed.


phil-davis:
Looks like this will pass. This is a demonstration PR only. It can be closed when CI finishes.

PR #621, opened 2021-02-14
[tests-only] Do not define SONAR_PULL_REQUEST env vars on master CI

phil-davis:
Issue https://github.com/owncloud/encryption/issues/238

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -1875,7 +1875,7 @@ def sonarAnalysis(ctx, phpVersion = '7.4'):
 				]
 			},
 			{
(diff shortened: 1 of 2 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/user_ldap/2983/16/7

The sonar-analysis pipeline ran just the sonarcloud-pr step, which is the correct expected behavior for a PR.

After merge, the sonarcloud-pr step should not run, but the sonarcloud-master step should run. And that should happen in nightly cron also. We will see after merge.

dpakach (approved):

PR #635, opened 2021-05-19
feat: add occ command to invalidate ldap caches

DeepDiver1975:


mmattel:
docs relevant. please file a docs issue when close to merge.
@EParzefall fyi

cdamken:
Is it possible to add the trigger for the search , that way the next user that will do the search and load the cache again won't have the bad experience to wait sometimes more than 2 minutes when they want to share. Additionally, if the command happens frequently as a cron job the bad experience from users will be incremented.

DeepDiver1975:
> Is it possible to add the trigger for the search , that way the next user that will do the search and load the cache again won't have the bad experience to wait sometimes more than 2 minutes when they want to share. Additionally, if the command happens frequently as a cron job the bad experience from users will be incremented.

The use case is  that an admin after adding a user to an ldap group uses this command (later press a button on the web ui) to invalidate this user's groups cache.
It is meant to be used as a scalpel and not as a hammer ;-)

DeepDiver1975:
> Is it possible to add the trigger for the search

search is not the point - the cached groups are used when the user loads the files - this is when group shares are resolved and shares are created in the file space. triggering this on each and every load would result in no cache being there. 
performance is :hankey: .... we don't want that

voroyam:
Steps to reproduce issue:
1. LDAP in test environment, make LDAP group available. Group should have at least one member.
2. Share file or folder with group from LDAP
3. Add another user to group from step 2 in LDAP backend (or remove the user from group)
4. Check with `occ group:list-members <group name>` the new user is showing (or no longer showing)
5. Check with `occ user:list-groups <user id>` the groups are showing for the user from step 3 (or no longer showing).
6. Check incoming shares in WebUI

---
Expected outcome:
Group share should immediately show for user from step 3 above (or disappear).

--- 
Actual behavior:
Group share is not shown (or still shown)


Then I tried `occ ldap:invalidate-cache` and `occ ldap:invalidate-cache oc3`

output was:

`No cache available - nothing to do ....`

I set the cache TTL for testing to 2 seconds, but then changed it back to 600. Tried the commands and the test again - same result.

DeepDiver1975:
that test server has no distributed cache enabled - command will have no impact ....

cdamken:
That's why I mean:

> The use case is that an admin after adding a user to an ldap group uses this command (later press a button on the web ui) to invalidate this user's groups cache.

The problem here is that the ownCloud admin is not necessarily the AD Admin and he doesn't know when it will be needed.

The cache will be loaded again when a search request is triggered by a user trying to share something new. There is where I say the User Experience will be: the ownCloud is too slow when you share. 

Well, that's my 2 cents, at least we have a command now, it's better than nothing ;)

voroyam:
After creating a new server with distributed cache enabled, I got the command to work, but the group share does not appear.

```
root@dmitry-oc-public:~# occ ldap:invalidate-cache
root@dmitry-oc-public:~# occ ldap:invalidate-cache oc3
 1/1 [============================] 100%root@dmitry-oc-public:~#
```

DeepDiver1975:
> The problem here is that the ownCloud admin is not necessarily the AD Admin and he doesn't know when it will be needed.

this was part of the specification -> please discuss in #concept on rc about this

cdamken:
> this was part of the specification -> please discuss in #concept on rc about this

Actually it is, if you read: https://github.com/owncloud/enterprise/issues/4533#issue-872529476

Point 3 : Add another user to group from step 2 in LDAP backend (or remove the user from group)

@voroyam and I just tested and removing the user and running the command, the is still in the group in ownCloud but not in the AD

jnweiger:
@DeepDiver1975 @cdamken @micbar  is this a quick one, that should this go into pending 0.15.3 release?

cdamken:
@jnweiger don't forget to include in the 0.15.3 release!

cdamken:
But afaik the patch didn't work and @voroyam and I tested but were not successful.

@DeepDiver1975  did you made some changes? I just saw your comment that you will be back on the 7th https://github.com/owncloud/enterprise/issues/4533#issuecomment-845096127

DeepDiver1975:
Since it did not work -> close 

Someone else took over .... Afaik JP ....

PR #636, opened 2021-05-31
Security: filter special characters from password field

C0rby:
Filter special characters like `\r\n` to prevent Server Side Request Forgery (SSRF) attacks.

### DIFF
--- a/lib/Configuration.php
+++ b/lib/Configuration.php
@@ -187,6 +187,7 @@ class Configuration {
 			$setMethod = 'setValue';
 			switch ($key) {
 				case 'ldapAgentPassword':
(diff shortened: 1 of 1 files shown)


jvillafanez:
Should we log something? "Password contains invalid chars that we've stripped" or similar?

C0rby:
I don't think it's necessary. 
With this PR we will filter out ASCII control characters. IMO it's only possible to set them into the password when you intercept the request and insert the special characters there. If you are using the UI it shouldn't be possible to insert the special characters since stuff like `\n` will be treated as just two characters and not the line feed character.

jvillafanez (approved):

voroyam:
@C0rby can you tell me what characters are now not allowed from the password filed? A customer just opened a ticket that the update breaks the login for some users with special characters in their passwords.

Issue #659, opened 2021-07-07
[QA] Frontend breaks with other auser auth apps


hurradieweltgehtunter:
oauth2 + user_ldap breaks tabs view

jnweiger:
![image](https://user-images.githubusercontent.com/1108546/124778390-3df0ed80-df41-11eb-885e-a239111bcbfc.png)

jnweiger:
Confirmed fixed in 0.15.4-rc1: First oauth2, then user_ldap. No overlap.

PR #660, opened 2021-07-07
fix display errors in combination with other apps

hurradieweltgehtunter:
#659


jnweiger (approved):
I'll call that v0.15.4-RC1 and test.

Issue #661, opened 2021-07-09
Update to 0.15.3 from 0.15.2 breaks user authentication


MascHman:
Hi,

we just upgraded from 0.15.2 to 0.15.3 and after successful upgrade our ldap users cannot login anymore. Using a fallback admin to log in, review the settings under User-Authentication -> LDAP reveals no errors whatsoever. All config tests are returning successful.

got some config logs of failed login attempts by my ldap user from that timeframe:

{"reqId":"onsmXM1u6idx5qqUaumP","level":2,"time":"2021-07-09T16:23:42+02:00","remoteAddr":"W.X.Y.Z","user":"--","app":"OCA\\User_LDAP\\User\\Manager","method":"POST","url":"\/index.php\/login?user=username","message":"No cached ldap result found for email_adress@domain.name"}
{"reqId":"onsmXM1u6idx5qqUaumP","level":2,"time":"2021-07-09T16:23:47+02:00","remoteAddr":"W.X.Y.Z","user":"--","app":"core","method":"POST","url":"\/index.php\/login?user=username","message":"Login failed: 'email_adress@domain.name' (Remote IP: 'W.X.Y.Z')"}
{"reqId":"onsmXM1u6idx5qqUaumP","level":2,"time":"2021-07-09T16:23:47+02:00","remoteAddr":"W.X.Y.Z","user":"--","app":"core","method":"POST","url":"\/index.php\/login?user=username","message":"Login failed: '11AE8D25-F1B0-49E4-9238-37837CAACD9E' (Remote IP: 'W.X.Y.Z')"}
{"reqId":"QPh32gmR0fPmJBSpSpoA","level":2,"time":"2021-07-09T16:24:12+02:00","remoteAddr":"W.X.Y.Z","user":"--","app":"core","method":"POST","url":"\/index.php\/login?user=email_adress@domain.name","message":"Login failed: 'username@domain.name' (Remote IP: 'W.X.Y.Z')"}

We're using an Active Directory as our LDAP backend and running on Windows 2016 Servers.
The Owncloud host itself is a standard Debian Linux (oldstable) running with PHP-7.4.21 , Apache 2.4.25

Is anyone else having this issue ?

I've downgraded our server back to 0.15.2 and everything is working again as before.

jvillafanez:
[...]

PR #663, opened 2021-07-14
Just strip "low" chars, but don't do anything else

jvillafanez:
For testing purposes:
```
// $b contains all ascii chars from 0-255
php > var_dump(filter_var($b, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW));
string(229) " !&#34;#$%&&#39;()*+,-./0123456789:;?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~��������������������������������������������������������������������������������������������������������������������������������"
php > var_dump(filter_var($b, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW));
string(224) " !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~��������������������������������������������������������������������������������������������������������������������������������"
```

### DIFF
--- a/lib/Configuration.php
+++ b/lib/Configuration.php
@@ -187,7 +187,7 @@ class Configuration {
 			$setMethod = 'setValue';
 			switch ($key) {
 				case 'ldapAgentPassword':
-					$val = \filter_var($val, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW);
+					$val = \filter_var($val, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW);
 					$setMethod = 'setRawValue';
 					break;
 				case 'homeFolderNamingRule':


jvillafanez:
Linking with https://github.com/owncloud/user_ldap/issues/661 and https://github.com/owncloud/user_ldap/issues/662

jnweiger (commented):
Interesting, that we filter characters at all. 
Should't we just keep any chars as they are? We don't control, what weird password patterns people have in their ldap. 
E.g. is a TAB character legal?

jvillafanez:
Original PR is https://github.com/owncloud/user_ldap/pull/636 . I'll leave the decision to the security experts.

JanAckermann:
>E.g. is a TAB character legal?
With this PR it is not possible to have a TAB in the password, this needs to be checked.

But keep nullbyte injections in mind, if you decide to remove the filter

cdamken:
> With this PR it is not possible to have a TAB in the password, this needs to be checked.

Is allowed to `type` tabs?

phil-davis:
It is difficult for a user to "insert" a tab into their password, because if they press the tab key in password input fields then the focus jumps out of the field to the next field. But it can be done if the user is "clever" and copies a tab character from some other place and pastes it into the password field.

Is there a "standard" somewhere that says what characters are supposed to be allowed/not allowed in a "standard" password?

jnweiger:
> Is there a "standard" somewhere that says what characters are supposed to be allowed/not allowed in a "standard" password?

It is most likely each LDAP Admin's discretion to allow / forbid / enforce certain character ranges. Not our call to divert.
As a user who has to type passwords while watched by strangers, I'd be happy for everything invisible.

IljaN:
> Is there a "standard" somewhere that says what characters are supposed to be allowed/not allowed in a "standard" password?

From what I know best practices are to allow everything as password. Tough I don't know if this rule includes non-printable characters. I guess not but needs to be researched ;)

phil-davis:
The original issue talked about making sure that carriage-return and line-feed characters were not "injected" into a password.
Maybe we should only strip those from the password string, and leave other "extra-special" characters like tab, backspace, escape, and the other "special binary values below ASCII %x20"

C0rby:
> > Is there a "standard" somewhere that says what characters are supposed to be allowed/not allowed in a "standard" password?
> 
> From what I know best practices are to allow everything as password. Tough I don't know if this rule includes non-printable characters. I guess not, but needs to be researched ;)

I think we can safely disallow/strip non printable characters from the password because I don't think there is a way a normal user would be able to get them into the password field.
Also I think the most used password mangers won't generate passwords with non-printable characters but I could be wrong on this point.

IljaN (approved):

voroyam:
> For testing purposes:
> 
> ```
> // $b contains all ascii chars from 0-255
> php > var_dump(filter_var($b, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW));
> string(229) " !&#34;#$%&&#39;()*+,-./0123456789:;?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~��������������������������������������������������������������������������������������������������������������������������������"
> php > var_dump(filter_var($b, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW));
> string(224) " !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~��������������������������������������������������������������������������������������������������������������������������������"
> ```

can you tell me what characters are allowed as per this PR and what are not allowed? I can't figure out what output is the legal and what is the illegal one.

voroyam:
simple language please :)

voroyam:
So everything that is not blacked out by my huge black bar is a legal password character? 

image

phil-davis:
https://www.asciitable.com/

Yes, the codes from 0 to 31 are "non-printable" things like null, backspace, tab, line-feed, carriage-return, escape and stranger codes for things that are never used by ordinary mortals. Those cannot be used in a password.

voroyam:
Okay, this is something my limited brain can understand. Thanks. 

Can you please tell me now the difference in the amount of legal characters between this PR and https://github.com/owncloud/user_ldap/pull/636?

phil-davis:
`FILTER_SANITIZE_STRING` in the other PR also removes things that look like HTML tags. Example:
```
$x = "<b>test</b>/best";
$y = \filter_var($x, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW);
echo $y;
```

The output is:
```
test/best
```

Notice that sequences like `<b>` and `</b>` are removed. So all sorts of characters can be removed if they occur between `<>`. But the `/b` in `/best` is not removed - it is not between `<>`.

So "it's complicated" - it is HTML tag patterns that were being removed. Passwords containing `<` and `>` would certainly have been changed.

PR #658, opened 2021-07-07
Merge back Release 0.15.4

jnweiger:
merge back to master, 
closes #605

PR #652, opened 2021-07-06
Only set USER_PREFKEY_FIRSTLOGIN to '1' on first login

JanAckermann:
This prevents firstLoginAccomplished log entries get written each time as a user logs in under certain circumstances (log level...).

PR #666, opened 2021-07-21
[0.15.3 branch] only set USER_PREFKEY_FIRSTLOGIN to '1' on first login

phil-davis:
GitHub is reporting a conflict in PR #658 for file `lib/User/Manager.php`

That was changed in master by PR #652 

Maybe if we backport that to the `release-0.15.3` branch then the conflict will go away?

Note: PR #652 will not be in either release 0.15.3 or release 0.15.4 - this is just trying to get rid of the reported conflict so that we can merge PR #652

### DIFF
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -450,12 +450,18 @@ class Manager {
 	 * @param string $uid
 	 */
 	public function markLogin($uid) {
-		$this->ocConfig->setUserValue(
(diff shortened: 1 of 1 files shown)


phil-davis:
I tested a merge of all this back to master in PR #667 and it looks good.

When CI  finishes here, I will merge this to `release-0.15.3`. Then CI will run for PR #658 and it should pass, and there should be no conflicts reported.

PR #669, opened 2021-07-22
Add function to check if connections are alive

JanAckermann:
links with https://github.com/owncloud/core/pull/39044 (although no hard dependency)

### DIFF
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -50,6 +50,7 @@ class User_Proxy extends Proxy implements
 	 */
 	private $backends = [];
 	private $refBackend = null;
+	private $serverConfigPrefixes = null;
 
 	/**
 	 * Constructor
@@ -57,6 +58,9 @@ class User_Proxy extends Proxy implements
 	 */
 	public function __construct(array $serverConfigPrefixes, ILDAPWrapper $ldap, IConfig $ocConfig) {
 		parent::__construct($ldap);
+
+		$this->serverConfigPrefixes = $serverConfigPrefixes;
+
 		foreach ($serverConfigPrefixes as $configPrefix) {
 			$this->backends[$configPrefix] =
 				new User_LDAP($ocConfig, $this->getAccess($configPrefix)->getUserManager());
@@ -64,6 +68,8 @@ class User_Proxy extends Proxy implements
 				$this->refBackend = &$this->backends[$configPrefix];
 			}
 		}
+
+
 	}
 
 	/**
@@ -362,5 +368,19 @@ class User_Proxy extends Proxy implements
 			$result = null;
 		}
 		return $result;
+	}
+
+	public function testConnection(){
+		foreach ($this->serverConfigPrefixes as $configPrefix) {
+			$userManager = $this->getAccess($configPrefix)->getUserManager();
+			$cr = $userManager->getConnection()->getConnectionResource();
+
+			$errorCode = $this->ldap->errno($cr);
+
+			if($errorCode === -1) {
+				throw new ServerNotAvailableException('Lost connection to LDAP server.');
+			}
+
+		}
 	}
 }
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -66,8 +66,6 @@ class User_Proxy extends Proxy implements
 				$this->refBackend = &$this->backends[$configPrefix];
 			}
 		}
-
-
 	}
 
 	/**
@@ -373,19 +371,15 @@ class User_Proxy extends Proxy implements
 	 *
 	 * @throws \Exception
 	 */
-	public function checkForConnectionErrors(){
+	public function checkForConnectionErrors() {
 		foreach ($this->serverConfigPrefixes as $configPrefix) {
 			$userManager = $this->getAccess($configPrefix)->getUserManager();
 			$cr = $userManager->getConnection()->getConnectionResource();
-
 			$errorCode = $this->ldap->errno($cr);
-			$error = $this->ldap->error($cr);
 
-			if($errorCode != ldap::LDAP_SUCCESS) {
-				throw new \Exception($error);
+			if ($errorCode != ldap::LDAP_SUCCESS) {
+				throw new \Exception($this->ldap->error($cr));
 			}
-
-
 		}
 	}
 }
--- a/lib/User_Proxy.php
+++ b/lib/User_Proxy.php
@@ -27,8 +27,6 @@
 
 namespace OCA\User_LDAP;
 
-use OC\ServerNotAvailableException;
-use OCA\User_LDAP\Exceptions\DoesNotExistOnLDAPException;
 use OCP\IConfig;
 use OCP\IUserBackend;
 use OCP\User\IProvidesEMailBackend;
@@ -250,7 +248,7 @@ class User_Proxy extends Proxy implements
 					return $backend->canChangeAvatar($uid);
 				}
 			}
-		} catch (\Exception $e) {
+		} catch (Exception $e) {
 			\OC::$server->getLogger()->debug($e->getMessage());
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


lib/User_Proxy.php
```diff
@@ -363,4 +369,18 @@ public function getUserName($uid) {
 		}
 		return $result;
 	}
+
+	public function testConnection(){
+		foreach ($this->serverConfigPrefixes as $configPrefix) {
```
JanAckermann, line 374: We should rename it to "checkForConnectionErrors" or "checkConnectionErrorsOccured"

lib/User_Proxy.php
```diff
@@ -244,7 +246,7 @@ public function canChangeAvatar($uid) {
 					return $backend->canChangeAvatar($uid);
 				}
 			}
-		} catch (\Exception $e) {
+		} catch (Exception $e) {
```
JammingBen, line 249:
```suggestion
		} catch (\Exception $e) {
```

jvillafanez:
There are a couple of problems with this approach:
* It seems better if the connection check is performed in the `User_LDAP` class. The proxy should just need to call the provided method in each configured backend.
* The errno check should be performed after each ldap call. There are 2 problematic scenarios with the current approach:
  * If a method of the `User_LDAP` class calls 2 different ldap functions in the same method, we'll only get the errno of the last function. This might or might not be a problem, depending on the specific method and how the error in handled inside that function.
  * Both user and group ldap classes share the same `Access` object, assuming the config prefix is the same. This could lead to problems if we accidentally mix the calls because we might get the error or success of the last operation which isn't the one we want to check. The process could be something like:
    1. Call user method -> fail
    2. Call group method -> success
    3. Check user method status -> show success instead of failure.

JammingBen:
Actually, each LDAP call gets wrapped via https://github.com/owncloud/user_ldap/blob/master/lib/LDAP.php#L286 already. In case the LDAP call returns `false`, we call `postFunctionCall()` which then throws an exception. The problem is, the LDAP method `search()` does not return false, but an empty array with false. That's why our wrapper does not throw an error like it usually does.

But again, changing this behavior might be critical as the search is not supposed to throw an exception (I guess?). So this approach is just a "hack" to propagate the error to the users iterator.

I still think that the cleanest approach is to dismiss all these approaches and just add a console output to the user sync command as described here at the end: https://github.com/owncloud/enterprise/issues/4642#issuecomment-885642746

jvillafanez:
The ldap wrapper can throw an exception, but that exception needs to be handled in the `User_LDAP` class (or somewhere in between) because core doesn't expect an exception (at least according to what is defined in the interface). In this case, we have to ensure the exception is propagated without issues up to the `User_LDAP` class. We need to pay special attention to cookie handling and any other state that could be left dangling.

Also note that the `User_Proxy` is expected to behave as a hub, so we should limit the methods we add there.

> I still think that the cleanest approach is to dismiss all these approaches and just add a console output to the user sync command as described here at the end: owncloud/enterprise#4642 (comment)

I think it's important to assume that the admin won't read the message. Of course, we have to show an error message, but it's also important to return an error code

JammingBen:
Okay I'm slowly losing the context here.

> The ldap wrapper can throw an exception, but that exception needs to be handled in the User_LDAP class (or somewhere in between) because core doesn't expect an exception (at least according to what is defined in the interface).

But the LDAP wrapper will never throw an exception during the user sync. It will only do so if a LDAP call returns false, which is never the case for the methods the user sync calls. Thus there is nothing to propagate.

> I think it's important to assume that the admin won't read the message. Of course, we have to show an error message, but it's also important to return an error code

If the admin does not read the console output of an executed command, he/she should not be executing any commands at all 😄  IMO that's an issue on the user's side, not on ours.

JanAckermann:
@JammingBen 
> If the admin does not read the console output of an executed command, he/she should not be executing any commands at all smile IMO that's an issue on the user's side, not on ours.

I think @jvillafanez is talking about return codes for the cron jobs, so admins can simply check if jobs failed. This point is legit and we can simply implement it, by catching the error in the command, print it return an error code.

jvillafanez:
From my point of view, the goal is to show a warning message in the console AND return a proper exit code.
In order to do so, we need to detect if a timeout has occurred and propagate that error up to the user:sync command. Assuming we have a proper way to detect the timeout, the question is how we're going to propagate the error. We've agreed that throwing an exception could be dangerous (we'd have to recheck all the connection code to ensure there is no problem)

For the code changes in this PR:
* The `User_Proxy` and `Group_Proxy` classes act as a hub of multiple `User_LDAP` and `Group_LDAP` objects respectively. Accessing the raw ldap connection isn't part of the responsibilities of these proxy classes. I think it's fine that the method is exposed in the proxy, but the action itself should be delegated.
* Both user and group proxies share the same connection (assuming the config prefix is the same). We need to be extremely careful because it could become a problem as seen in https://github.com/owncloud/user_ldap/pull/669#issuecomment-886428414 

I'm not sure if we're focusing too much on fixing the `user:sync` command disregarding the user_ldap app. I expect, whatever the app provides, it will be complete. In this case, the group proxy should have the same mechanism in order to keep consistency.

> First we said, don't touch legacy code and now we should touch it?

I'm not saying you mustn't touch legacy code. I'm saying that touching legacy code will imply A LOT of testing to ensure we don't break anything. This goes far from just checking that the code fixes the problem; we have to ensure there is no side effect in other calls (what happens to the pagination cookie if an exception is thrown?), and we have to test with multiple LDAP providers.
This is why we might want to look into lighter alternatives.

JammingBen:
> Assuming we have a proper way to detect the timeout, the question is how we're going to propagate the error.

@jvillafanez I'm running out of ideas to achieve this. Maybe we could:

* Move the `checkForConnectionErrors()` method from the `User_Proxy` to the `Connection` class instead. This way it is clear that we're looking for errors in the connection itself, not just for users or groups.
* Instead of throwing an exception, we return an array of all errors that occurred. The caller can then decide what to do about it.
* Implement a `getUserManagers()` method in the `User_Proxy` class to ensure that other services can get these managers from the outside. As these managers provide direct access to the LDAP connection, `checkForConnectionErrors()` can be called as well from the outside.

Once again, `checkForConnectionErrors()` is a somewhat "hacky" way to retrieve errors that don't already get thrown via https://github.com/owncloud/user_ldap/blob/enterprise/issues/4642/lib/LDAP.php#L317. Which should be the case for almost all LDAP calls, except a few exceptions (like the one in this case).

jvillafanez:
> Move the checkForConnectionErrors() method from the User_Proxy to the Connection class instead. This way it is clear that we're looking for errors in the connection itself, not just for users or groups.

Since both the `User_LDAP` and `Group_LDAP` will share the same connection, we should store the last connection error there. This means that both the `User_LDAP` and `Group_LDAP` must check for connection errors automatically on each operation in order to store the error (if any).
The `checkConnectionErrors` in the proxies will delegate to the respective `User_LDAP` and `Group_LDAP`, which will return the last stored error.

> Instead of throwing an exception, we return an array of all errors that occurred. The caller can then decide what to do about it (e.g throw exception, log exception and return code, ...)

I don't see a big difference with the previous item. Core will still need to call to an optional method in the backend. The only difference is that, instead of returning the last error, it will return an array with multiple errors. This seems to be more complex to handle, and we need to be very careful about how the error info is stored (there might be references to objects that will be difficult to free, and might consume a lot of memory).

> Implement a getUserManagers() method in the User_Proxy class to ensure that other services can get these managers from the outside. As these managers provide direct access to the LDAP connection, checkForConnectionErrors() can be called as well from the outside.

Exposing the native LDAP connection is a bad idea. In fact, the native LDAP connection shouldn't be exposed at all by the ldap wrapper, but that's a terrible design decision we have to live with. Note that anyone could access the native resource and close the connection by itself, causing a lot of problems that will be extremely hard to debug, this is why we mustn't expose the native resource outside of the app.

> Once again, checkForConnectionErrors() is a somewhat "hacky" way to retrieve errors that don't already get thrown via https://github.com/owncloud/user_ldap/blob/enterprise/issues/4642/lib/LDAP.php#L317. Which should be the case for almost all LDAP calls, except a few exceptions (like the one in this case).

I wouldn't mind to throw an exception from the very bottom (probably the ldap wrapper), but the problem is that we need to transfer the exception from the ldap app to core. Normally, we need to adjust the interface so it's allowed to throw an exception in all (?) of the public methods of the `UserInterface`, but we'll have to adjust also core code and **any other app** using the `UserInterface`. I think we'd need some core releases before release the fix in the LDAP app in order to stabilize  and adjust the rest of the code (and give enough time to 3rd party apps to adjust their code too). You can see that, even the solution seems simple, there is a lot of work involved in order to prevent breaking things.
In the end, we'll have to rely on a new optional `checkConnectionErrors` method so core can check for errors without disturbing the rest of the apps.

I completely agree it isn't the best solution, but otherwise, it seems we'd need to change the architecture.

JammingBen:
> This means that both the User_LDAP and Group_LDAP must check for connection errors automatically on each operation in order to store the error (if any).

Maybe I'm on the wrong track, but this is what I'm trying to tell you: Exactly this is already happening in https://github.com/owncloud/user_ldap/blob/master/lib/LDAP.php. Each LDAP call gets wrapped and throws an exception if an error pops up. With the search call being the (only?) exception here, because it returns the error in an array, which the wrapper does not detect. But I don't want to change this, because maybe this is by design - I don't know 😄  On the other hand, I also don't want to implement another architecture for error handling here.

JammingBen:
This is a "won't fix" for now: https://github.com/owncloud/enterprise/issues/4642#issuecomment-986586722

PR #670, opened 2021-07-22
Add a command to invalidate the LDAP cache

jvillafanez:
Link to https://github.com/owncloud/enterprise/issues/4533

The command will remove ALL the cache entries associated to the LDAP app.

PR #671, opened 2021-07-30
Adjust command description

jvillafanez:
The invalidate-cache command will wipe out the whole cache. There isn't any user-based action


phil-davis (approved):
Now the text just repeats the words of the command itself, in a different order. But that is what the command does - writing a whole explanation of "what is a cache", "what is the LDAP cache" etc is not possible in a 1-sentence description!

mmattel:
Corrects the wrong text in: https://github.com/owncloud/user_ldap/pull/670 (Add a command to invalidate the LDAP cache)

PR #672, opened 2021-08-03
fix user group selection layout

hurradieweltgehtunter:
Fix layout of user group selection boxes


hurradieweltgehtunter:
Before:
grafik


After:
grafik

voroyam:
works for me

image

hurradieweltgehtunter:
Let me know if the red warning triangles should stay or not

voroyam (approved):

voroyam:
Approve from me. The PR does what it says it does. The Warning signs can be handled with a different PR after we come to a decision.

mmattel:
I agree with @voroyam that the red warning triagles are more a _dont touch it_ thing. Maybe just make them orange would relax the optics. (adding myself here because I have do document the new layout, making consecutive screenshot changes with minimal changes is a lot of effort)

voroyam:
> I agree with @voroyam that the red warning triagles are more a _dont touch it_ thing. Maybe just make them orange would relax the optics. (adding myself here because I have do document the new layout, making consecutive screenshots of the same with minimal changes in the optics is a lot of effort)

imho the symbol it self is the issue here, the color is irrelevant. It demands attention.

as stated before, I don't think we need further efforts to help the users, what we already have is enough.

jnweiger:
> the red warning triagles are more a _dont touch it_ thing

That does not work. I never succeeded configuring an LDAP server without touching at least Advanced.
Internal state machine of the ldap config is too broken, to allow a simple walk through the tabs from left to right. 
It is always a back and forth, trying different things. The entire interface is experts only.

(But all that is offtopic here)

jnweiger:
@voroyam @hurradieweltgehtunter I need help, how to get to that 
"Available Groups" widget in the first place.

* With openldap, where memberof is not supported, The groups widgets in the user tab always remain inactive
image

----
* With AD where memberof is supported, I can select groups in the user tab, but the widget looks totally different than what you get:
image

mmattel:
It is interesting that the widget is inactive when using openldap and active when using AD. This is docs relevant, pls file a docs issue.

jnweiger:
I solved the puzzle:
a) When the LDAP server does not support MemberOf, then it is inactiv. 
b) When the admin forgot to click on "Edit LDAP Query" (which is actually the "Mode Switch" Button) then it is inactive.
c) When there are 40 or less groups in the server, then a dropdown with checkboxes is shown.
d) when there are 41 or more groups in the server, then the tow column widgets with [<], [>] arrows is shown.

jnweiger:
Confirmed fixed in 10.9.0 RC2 (the part that was fixed by Flo)

PR #677, opened 2021-08-17
do not skip user on OOBE

mrow4a:
- [ ] add unit tests

related to https://github.com/owncloud/enterprise/issues/4686#issue-951464336 .

Before this change we silently skipped the user with collision mapping, and just adding exception in the log. However, this could have led to situation when administrator could misconfigure his instance and user sync successfully works on misconfigured instance

before:
Zrzut ekranu 2021-08-17 o 10 55 26

after:
Zrzut ekranu 2021-08-17 o 10 54 50

### DIFF
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -489,6 +489,7 @@ class Manager {
 	 * @param integer $limit
 	 * @param integer $offset
 	 * @return string[] an array of all uids
+	 * @throws OutOfBoundsException
 	 */
 	public function getUsers($search = '', $limit = 10, $offset = 0) {
 		$search = $this->access->escapeFilterPart($search, true);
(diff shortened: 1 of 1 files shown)


mrow4a:
@butonic any issue you potentially see with this? except for the fact that now sync will break instead of just skipping the user with collision.

jvillafanez:
The problem I see is that the exception will bubble up to the `User_LDAP` class, which implements the `UserInterface`. The interface doesn't say anything about throwing an exception in the `getUsers` method, so we're breaking the interface.

While it could fix the issue, it could also break other things because such exception isn't expected to be thrown.

If we have to change the interface, we'll have to plan what changes are needed and adjust core (and maybe other apps) to those changes.

mrow4a:
@jvillafanez hmm ok true.. probably https://github.com/owncloud/core/blob/master/lib/public/UserInterface.php#L61-L70 would need to be adjusted to throw exception. 

let me think on workaround for this just for that user:sync command..

mrow4a:
@jvillafanez I think it just needs the documentation then.. it is thrown in misconfiguration corner-case

mrow4a:
https://github.com/owncloud/docs/pull/4017

PR #678, opened 2021-09-01
change color of warning sign

hurradieweltgehtunter:
In adtion to #672 

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased] - XXXX-XX-XX
 
+### Changed
+- Change color of warning sign for expert settings [#678](https://github.com/owncloud/user_ldap/pull/678)
 
 ## [0.15.4] - 2021-07-13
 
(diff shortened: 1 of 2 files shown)


mmattel (approved):
LGTM
The use of orange makes it better differenciateable from the red error block.

jnweiger:
As an average admin, I would not know the specific use of orange here. 

The meaning here is "Everything is currently fine, but be cautious, when you enter these tabs"
It is easily misunderstood as "Something needs your attention in this tab"

WND e.g. also uses orange in the admin interface -- there it means again something different (and non-guessable)

Sorry for lobbying against this feature. I still fail to see a usecase that would outweight the possible confusion, that I can imagine.

voroyam:
> As an average admin, I would not know the specific use of orange here. 
> 
> 
> 
> The meaning here is "Everything is currently fine, but be cautious, when you enter these tabs"
> 
> It is easily misunderstood as "Something needs your attention in this tab"
> 
> 
> 
> WND e.g. also uses orange in the admin interface -- there it means again something different (and non-guessable)
> 
> 
> 
> Sorry for lobbying against this feature. I still fail to see a usecase that would outweight the possible confusion, that I can imagine.

Total agree.

hurradieweltgehtunter:
I happily do another change if you let me know how it should be

voroyam:
My view has not changed - remove the warning sign all together. The reasoning behind my view I described in a comment above.

PR #709, opened 2022-01-11
[tests-only] [full-ci] Test against 10.9.1RC2

phil-davis:
10.9.1RC2 tarballs are now available, so use them.

Note: many of the pipelines run in nightly CI. So this needs to be merged, and we will get results overnight.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -101,7 +101,7 @@ config = {
                 "oracle",
             ],
             "servers": [
-                "10.9.1RC1",
+                "10.9.1RC2",
             ],
             "phpVersions": [
                 "7.3",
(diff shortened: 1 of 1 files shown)


pascalwengerter (approved):
Is there a use case for having all of them defined individually? Otherwise I'd recommend a single constant, that makes changes way easier and less prone for "Uh I forgot one pipeline" 😇 but LGTM for now

phil-davis:
> Is there a use case for having all of them defined individually? Otherwise I'd recommend a single constant, that makes changes way easier and less prone for "Uh I forgot one pipeline" innocent but LGTM for now

yes, there are apps like this that have various combinations of pipelines that do different test suite... We could reduce the duplication of references to core tarball versions.

PR #711, opened 2022-01-17
Decode binary guid

IljaN:
With Oracle eDirectory we sometimes get a GUID as binaray where our code would normally expect a string.

https://github.com/owncloud/enterprise/issues/4686

- [x] Unit-Tests

### DIFF
--- a/lib/User/UserEntry.php
+++ b/lib/User/UserEntry.php
@@ -365,6 +365,11 @@ class UserEntry {
 			if ($trim) {
 				$value = \trim($value);
 			}
+
+			if ($attributeName === 'objectguid' || $attributeName === 'guid') {
+				$value = Access::binGUID2str($value);
+			}
+
 			if ($value === '') {
 				return $default;
 			}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 - Change color of warning sign for expert settings [#678](https://github.com/owncloud/user_ldap/pull/678)
+- Decode binary GUID where we normally expected string (eDirectory) [#711](https://github.com/owncloud/user_ldap/pull/711)
 
 ## [0.15.4] - 2021-07-13
 
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -1862,10 +1862,23 @@ class Access implements IUserTools {
 	 * {@link http://www.php.net/manual/en/function.ldap-get-values-len.php#73198 The PHP ldap_get_values_lan doc comments}
 	 *
 	 * @param string $binGuid the ObjectGUID / GUID in it's binary form as retrieved from Microsoft AD / Novell eDirectory
+	 *                        If you pass an already decoded GUID as string, it will be returned as is.
 	 * @return string
 	 * @throws \OutOfBoundsException
 	 */
 	public static function binGUID2str($binGuid) {
+		$guidLength = \strlen($binGuid);
+
+		// The guid should have 16 byte when binary and 36 byte when string (including '-' characters)
+		if (($guidLength !== 16) && ($guidLength !== 36)) {
+			throw new \OutOfBoundsException(\sprintf('Invalid GUID with length %d received: <%X>', $guidLength, $binGuid));
+		}
+
+		// If we get a guid in string form we simply return it to prevent double decoding
+		if ($guidLength === 36) {
+			return $binGuid;
+		}
+
 		// V = unsigned long (always 32 bit, little endian byte order)
 		// v = unsigned short (always 16 bit, little endian byte order)
 		// n = unsigned short (always 16 bit, big endian byte order)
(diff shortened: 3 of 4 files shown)


IljaN:
@jvillafanez Is it possible to detect beforehand if the guid is a binary? The decoding function expects a binary and errors if it is not. Maybe we can make it transparent?

IljaN:
Would this be the correct check?
https://github.com/owncloud/user_ldap/blob/ebea8a6f7f9c2026085308e834e308d93f1acfb4/lib/Access.php#L1874-L1879

jvillafanez:
I don't think we can do that reliably. Even if a string is utf8-encoded, it could also be considered as a binary string which happens to have a particular pattern.
It seems better to rely on the attribute definition and assume that the LDAP server will comply with it. The main problem is to find a reliable source of information

jvillafanez:
> Would this be the correct check?

Not really. The check works only if there isn't enough input
```
php > $binGuid = '1234567';
php > $unpacked = \unpack('Va/v2b/n2c/Nd', $binGuid);
PHP Warning:  unpack(): Type v: not enough input, need 2, have 1 in php shell code on line 1
php > echo \sprintf('%08X-%04X-%04X-%04X-%04X%08X', $unpacked['a'], $unpacked['b1'], $unpacked['b2'], $unpacked['c1'], $unpacked['c2'], $unpacked['d']);
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
PHP Notice:  Trying to access array offset on value of type bool in php shell code on line 1
00000000-0000-0000-0000-000000000000
```
in any other case, it will go through even if the content isn't a guid, such as a random string, a "X-Y1-Y2-Y3-Z" string (as it is expected to return) and any binary content
```
php > $binGuid = file_get_contents('/usr/bin/ls');
php > $unpacked = \unpack('Va/v2b/n2c/Nd', $binGuid);
php > echo \sprintf('%08X-%04X-%04X-%04X-%04X%08X', $unpacked['a'], $unpacked['b1'], $unpacked['b2'], $unpacked['c1'], $unpacked['c2'], $unpacked['d']);
464C457F-0102-0001-0000-000000000000
```

IljaN:
Idea would be to create a isBinary() function and check beforehand, but as you said there might be false-positives.... :thinking:

IljaN:
We don't know if the GUID is always binary in this case or just "sometimes"

jvillafanez:
According to https://ldapwiki.com/wiki/2.16.840.1.113719.1.1.4.1.501 , the guid should have 16 bytes.
We could try:
* If the string contains exactly 16 bytes, it's a binary string and we need to decode it.
* If the string contains exactly 36 bytes, it's our string representation, which we should consider valid (it might be a different thing, but I don't think we'll be able to distinguish just with the string itself). Nothing to do
* In any other case, throw an error.

lib/User/UserEntry.php
```diff
@@ -365,6 +365,20 @@ private function getAttributeValue($attributeName, $default = null, $trim = true
...
+
+			if ($attributeName === 'objectguid' || $attributeName === 'guid') {
+				$guidLength = \strlen($value);
+				// Sometimes we might get a binary guid (16 byte), so we decode it
+				if ($guidLength === 16) {
+					$value = Access::binGUID2str($value);
+				}
+
+				// GUID should be 16 bytes (binary) or 32 bytes decoded to string
+				if (($guidLength !== 32) && ($guidLength !== 16)) {
```
jvillafanez, line 377: It should be 36 not 32; we need to count the "-" chars too.

IljaN: :+1:

IljaN:
Spooky 
```
9) OCA\User_LDAP\Tests\User\UserEntryTest::testGetUUIDIsAuto with data set #4 ('guid', Binary String: 0x81c9534cba5dd...bd48a51, '4C53C981-5DBA-11D9-89A2-890B9BD48A51')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'4C53C981-5DBA-11D9-89A2-890B9BD48A51'
+'33354334-3943-3138-2D35-4442412D3131'
```

IljaN:
Hmhh, we double decode, hitting this codepath:
https://github.com/owncloud/user_ldap/blob/4a27401e5bfbe4ba073c34cf0e7288f99129e824/lib/User/UserEntry.php#L167

IljaN:
Tests should pass now. @jvillafanez Should we move those checks[1][2] in to the binGUID2str function so that it returns the passed value if a string with length 36 is passed to it?

[1] https://github.com/owncloud/user_ldap/blob/1fb0d70900031c0022f5c6fe51233ec0fade70aa/lib/User/UserEntry.php#L168
[2] https://github.com/owncloud/user_ldap/blob/1fb0d70900031c0022f5c6fe51233ec0fade70aa/lib/User/UserEntry.php#L374-L382

jvillafanez:
>  @jvillafanez Should we move those checks[1][2] in to the binGUID2str function so that it returns the passed value if a string with length 36 is passed to it?

We'd need to adjust the PHPDoc of the method, but it make sense. :+1:

jvillafanez (approved):

Issue #729, opened 2022-03-23
Acceptance tests failing


phil-davis:
Last night all nightly acceptance test pipelines failed with a problem in `BeforeScenario # FeatureContext::before()`

For example https://drone.owncloud.com/owncloud/user_ldap/3872/24/12

`Empty password not allowed - see allowEmptyPassword option. (Laminas\Ldap\Exception\LdapException)`

core PR https://github.com/owncloud/core/pull/39893 was merged yesterday. That made changes to LDAP things in acceptance tests - so the code changes there should be examined.

PR #731, opened 2022-03-23
[tests-only] [ full-ci] Fix typos in README

phil-davis:
The real purpose for doing this PR is to see if issue #729 is fixed.

It was still broken. I added a commit to test against core branch `adjust-ldap-test-code` - that looks like it will work.


phil-davis:
passes - great

PR #734, opened 2022-04-13
Rise default network timeout to 15 secs

jvillafanez:
2 seconds might be a too small timeout for average servers. 15 seconds should be enough.

For upgrades, it's expected that ownCloud will use the default value (previously 2 secs, now 15 secs) unless the admin changes it later. We might need a migration to explicitly set the value to "-1" to disable the timeout in order to prevent issues (**to be checked what's the actual value to disable the timeout**)

Note that the setting can still be adjusted manually if needed.


ownclouders:
:boom: Acceptance tests pipeline <strong>webUIUsrLDAPS-master-chrome-mysql8.0-php7.3</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/user_ldap/3908/66

pako81:
do we want to have this in 0.170?

jnweiger:
I'd like to include this in 0.17.0 -- I don't understand the implications of JPs concerns about migrations here.

pako81:
what about the case the admin changed the value before the upgrade? I guess it should not be set back to the new default (15) as it should be already written in the oc_appconfig table.

jvillafanez:
Yes, it should work like that.

pako81:
So maybe we can think about just documenting the new default so that admins are informed about this change. That should be enough.

pako81 (approved):

PR #748, opened 2022-07-22
[full-ci] Expose group's displayname

jvillafanez:
Expose the group's displayname and store the group id in the "owncloud_name" column of the "oc_ldap_group_mapping" table.

**For new installations**
By default, the "owncloud_name" will likely be entryuuid (at least checked with openldap, need to test with more providers). This will guarantee uniqueness. The default displayname will still be the cn.

TODO (in this PR):
* [x] Configurable ldap attribute for the "owncloud_name" in the expert tab of the wizard. The restrictions will be the same as the username: immutable, unique and only `a-zA-Z0-9+_.@-` chars allowed
* [x] Migration for updates. We'll force the configured displayname to be used instead of the default one.
* [x] Caching. For now, the displayname isn't cached. We probably should cache it at least to keep consistency with the rest of the app. Right now, it seems the updating the displayname in the LDAP automatically updates the name shown in ownCloud, but this might be the only case, so we might want consistency. Moreover, caching will reduce the hits the ldap server needs to take.

Ref: https://github.com/owncloud/enterprise/issues/5071


ownclouders:
:boom: Acceptance tests pipeline <strong>apiProvLDAPS-latest-mysql8.0-php7.3</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/user_ldap/4045/40

jvillafanez:
We need to decide if we raise the minimum version or not. The code is technically compatible with previous core versions, but by default it will show the uuid of the group which will look ugly. We'll need to verify, but the cases should be:
* OC 10.10- and user_ldap 0.16.0 => current behavior, it shows the cn
* OC 10.10- and user_ldap 0.17.0 (fresh) => it shows the uuid by default unless other attribute was configured. It can't be changed after it's set up.
* OC 10.10- and user_ldap 0.17.0 (updated) => it shows the cn due to migration, in order to keep the behavior from 0.16.0
* OC 10.11+ and user_ldap 0.16.0 => it shows the cn
* OC 10.11+ and user_ldap 0.17.0 (fresh) => it shows the cn. The internal groupname will be the uuid unless configured otherwise
* OC 10.11+ and user_ldap 0.17.0 (updated) => it shows the cn. The internal groupname will be the cn due to the migration. This can't be changed. The group displayname can be changed if needed.

The cases to check are the second and third. Second case looks ugly, but it should work. Note that OC 10.10 doesn't show the displayname of the group in most cases.

In addition, we've added caching for the group details. You might need to clear the cache (if it's being used) for the changes to appear.

jvillafanez:
@phil-davis I think we'll need to adjust the tests here. The provisioning API uses the gid, which is mapped to the entryuuid by default in this PR.
The alternative is to set the new expert groupname attribute to use the old cn. That should make the tests pass.

js/wizard/wizardTabExpert.js
```diff
@@ -73,6 +77,15 @@ OCA = OCA || {};
 			this.setElementValue(this.managedItems.ldap_expert_username_attr.$element, attribute);
 		},
 
+		/**
+		 * sets the attribute to be used to create an ownCloud ID (username)
```
phil-davis, line 81: comment needs editing

templates/settings.php
```diff
@@ -290,6 +290,14 @@
 					<input type="text" id="ldap_expert_username_attr" name="ldap_expert_username_attr" data-default="<?php p($_['ldap_expert_username_attr_default']); ?>" />
 				</div>
 			</section>
+			<section>
+				<h3><?php p($l->t('Internal Groupname')); ?></h3>
+				<p><?php p($l->t('The internal groupname is used to uniquely identify the group. It has the same restrictions as the internal username, in particular, the group name must be immutable and unique. By default, the UUID will be used. This internal groupname won\'t likely by visible because a displayname attribute is intended to be used to show the group.')); ?></p>
```
phil-davis, line 295: "It has the same restrictions as the internal username" - but probably a larger set of characters will work.
Do we try to say what characters are allowed?

jvillafanez: The internal username has `a-zA-Z0-9+_.@-` explicitly written, so those should be ok.
As you've said, a larger char set will likely work, but to keep it on the safe side, I'd rather keep the char set small so we don't face issues with weird chars.
The main idea is that it will behave the same way as the internal username, so it makes sense to have the same restrictions. We might want to include the whitespace explicitly in the char set because it's a common char, but other than that, I think it's fine as it is.

phil-davis:
The core daily-master-qa tarball was a week out of date. That was noticed today and it has been recreated a few hours ago. So it now has the latest code from core master, including from PR https://github.com/owncloud/core/pull/40229

So I restarted CI just now. That will get a fresh result, and then I will look at what tests fail and how we can fix them.

phil-davis:
@jvillafanez see my comment https://github.com/owncloud/user_ldap/pull/749#issuecomment-1202733705

There is a problem with group names that have "special characters" in them. Probably the group id or display name needs some escaping somewhere in the code - can you have a look?

I adjusted the test settings in commit https://github.com/owncloud/user_ldap/pull/748/commits/8ba4029de5fa31c164edfab089c8e27f91e66fe6 - now it defines `ldapExpertGroupnameAttr` in the config. And most of the test pass, which is good. I think that in CI the LDAP app was being installed, and `occ upgrade` happened that runs the migration. But that happens before the settings are created - the CI creates a whole set of settings with `tests/acceptance/setConfig.sh` and now those have to include `ldapExpertGroupnameAttr`

For installation that are already set up, the migration code should work nicely.

For new installations, `ldapExpertGroupnameAttr` has to be set. Maybe it would be a bit easier if, when it is not set, it defaults to using the group id as the group display name. That would avoid making everybody have to set `ldapExpertGroupnameAttr`

When we get all the tests passing, then I can sort out some scenarios in a separate test suite that has the group id and group display name different.

jvillafanez:
> For new installations, ldapExpertGroupnameAttr has to be set.

It should be already defaulting to the entryuuid (or any other uuid attribute) of the group. The behavior should be the same as with the username.

phil-davis:
I pushed commit 8b21e28ce57a2caef1035a3b79fc6ce557197ccc to my test branch.
That avoids a problem that I noticed:
`Trying to access array offset on value of type bool at /home/phil/git/owncloud/core/apps-external/user_ldap/lib/Group_LDAP.php#1022`

That seems to happen because `getGroupDetails` can be called even when the group does not exist - see my comments in the "Fixme" of that commit. Can you have a look at those code paths and make an appropriate fix to this PR.

jvillafanez:
> I pushed commit https://github.com/owncloud/user_ldap/commit/8b21e28ce57a2caef1035a3b79fc6ce557197ccc to my test branch.

It might be better to return null if the displayname isn't found and let core handle it.
```
if (!\is_array($displayname)) {
  // displayname not found
  return null;
}
```

phil-davis:
> It might be better to return null if the displayname isn't found and let core handle it.

core https://github.com/owncloud/core/pull/40261 refactored `getGroupObject` so that it should only call into `getGroupDetails` when the group exists in the backend. So that should prevent the case where you commented `FIXME: It seems local groups also end up going through here...`

And some of my comments in https://github.com/owncloud/user_ldap/commit/8b21e28ce57a2caef1035a3b79fc6ce557197ccc are also now out-of-date.

So yes, add your suggestion to return null if the display name is not found, if you like.

Then I think that this PR is ready.

jnweiger:
Note to self: There seems to be an unreleased 0.17.0 in circulation, so the next release should probably be 0.18.0 to be on the safe side.

jvillafanez:
We might need to delay this PR until 10.11.1 version (or the next one) is released due to https://github.com/owncloud/core/pull/40412

IljaN (approved):

lib/Group_LDAP.php
```diff
@@ -1001,6 +1001,33 @@ public function groupExists($gid) {
 		return true;
 	}
 
+	public function getGroupDetails($gid) {
```
mrow4a, line 1004: doc missing, and in fact I would appreciate the explanation.

lib/Access.php
```diff
@@ -530,38 +529,36 @@ public function groupsMatchFilter($groupDNs) {
...
+	public function dn2ocname($fdn, $isUser = true) {
 		if ($isUser) {
 			$mapper = $this->getUserMapper();
-			$displayNameAttribute = $this->connection->ldapUserDisplayName;
+			$nameAttribute = (string)$this->connection->ldapExpertUsernameAttr;
 		} else {
 			$mapper = $this->getGroupMapper();
-			$displayNameAttribute = $this->connection->ldapGroupDisplayName;
+			$nameAttribute = (string)$this->connection->ldapExpertGroupnameAttr;
 		}
```
mrow4a, line 562: add comment why not using ldapUserDisplayName/ldapGroupDisplayName

mrow4a (changes requested):

lib/Group_LDAP.php
jvillafanez: Not documented in the abstract class nor in the interface :facepalm: 
I'll open a ticket in core to document this where it should be. I'm adding the docs here too.

lib/Access.php
jvillafanez: I don't think there is any need. The internal ownCloud name should be internal, that's what the username / groupname attribute is for. A displayname is meant to be displayed, so using it for something internal doesn't make too much sense.

For the code changes, the user displayname was requested but not used (the username attribute was used instead), and the group displayname doesn't make sense as internal name as said.

mrow4a: but this does not brake anything right ?

jvillafanez: I don't think so. There is a migration in place to copy the old displayname attribute (which usually is the "cn") into the new groupname attribute, so there won't be any noticeable change after the upgrade.

For new installation, the groupname will default to the uuid, which should be the entryuuid or entryguid (maybe others, depending on the provider), but it can be adjusted if needed in the wizard.

pako81:
Is there anything open here? /cc @enbrnz

jvillafanez:
Maybe QA can confirm if we should rise the minimum version to 10.12 due to https://github.com/owncloud/core/pull/40412 . Other than that, I think @mrow4a needs to give the green light

jvillafanez:
I think https://github.com/owncloud/user_ldap/pull/748#issuecomment-1193762273 still applies.
Basically, problematic cases seems to be with core 10.11 and earlier, and ldap with this PR, in a fresh installation. The problem with a fresh 10.10 core (and earlier) is explained in the linked comment. A fresh 10.11 will be affected by https://github.com/owncloud/core/pull/40412.

Options are:
* Document the limitation without raising the minimum version. If people don't want problems, they'll likely need to change the new groupname id attribute, which might not happen.
* Rise the minimum version to 10.12 (or 10.11.1 or whatever next version is) so people won't be affected.

Note that the problem only happens in new installations. There is a migration in this PR in so the updates will keep the current behavior.

mrow4a:
@jvillafanez @cdamken I think the option agreed was no. 1 -> **release new user_ldap that includes this PR AFTER oc10.12 is released.** 

If there is corner case customer (with new instalation on oc10.11) like you explained, we would just recommend to upgrade to oc10.12 to have this feature on new user_ldap working. I think raising minimal version is a bit too much here.

phil-davis:
@jvillafanez GitHub is reporting conflicts in `appinfo/info.xml`
Can you rebase some time? (it should be very easy to resolve)

jvillafanez:
Conflicts in the info.xml were due to the different app version. I've set the version to 0.17.0 because we'll have to run the migration included in the PR and I'm not sure if it would run with just a minor version update.

Issue #759, opened 2022-11-02
AD objectsid attribute requires special handling


jvillafanez:
The objectsid attribute from AD could be used as uuid attribute in the expert tab of the wizard. However, the attribute is binary and it's giving problems

The binary attribute should be transformed so it can be stored in the DB and shown in the logs.

jvillafanez:
@d7oc

d7oc:
@jvillafanez I think we can close this issue as well.

PR #760, opened 2022-11-02
Binary converter

jvillafanez:
Refactor code to allow conversion of binary attributes.

Previously only the "objectguid" and "guid" were being handled. Other attributes such as "objectsid" weren't. The refactor makes adding support for more binary attributes easier just by creating the converter and registering it in the converter hub.
For now, the 3 mentioned binary attributes are supported and they'll show the expected string representation. More can be added if needed.

Ref https://github.com/owncloud/user_ldap/issues/759

### DIFF
--- a/lib/Attributes/ConverterHub.php
+++ b/lib/Attributes/ConverterHub.php
@@ -123,7 +123,7 @@ class ConverterHub {
 	 * @param string $attr the attribute associated to the string representation
 	 * @param string $strValue the string representation of the value of
 	 * the attribute
-	 * @param string a string suitable to be used as filter
+	 * @return string a string suitable to be used as filter
 	 * @throws ConverterException if no converter is registered, or the
 	 * associated converter can't deal with the input.
 	 */
--- a/lib/Attributes/ConverterHub.php
+++ b/lib/Attributes/ConverterHub.php
@@ -1,0 +1,138 @@
+<?php
+/**
+ * @copyright Copyright (c) 2022, ownCloud GmbH.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


lib/User/UserEntry.php
```diff
@@ -163,9 +164,6 @@ public function getUUID() {
 				$this->connection->ldapExpertUUIDUserAttr = $uuidAttribute;
 				$this->connection->saveConfiguration(); // FIXME should not be done here. Move to wizard?
 			}
-			if ($uuidAttribute === 'objectguid' || $uuidAttribute === 'guid') {
```
jvillafanez, line 166: This shouldn't be needed since the value should be transformed in the `getAttributeValue` call. The change in the `getAttributeValue` was added in https://github.com/owncloud/user_ldap/pull/711 due to some problems

mrow4a:
wow cool, I remember we once already had a problem with binary field.. would have a detailed look later tomorrow

butonic (approved):

lib/Attributes/ConverterHub.php
```diff
@@ -0,0 +1,138 @@
...
+	/**
+	 * Get the default converted hub. If no hub has been created yet,
+	 * or the $fresh parameter is true, a new ConverterHub will be
+	 * created and set as default, the `registerDefaultConverters`
+	 * method will be called on it, and finally it will be returned.
+	 * @param bool $fresh whether a new instance should be created and set
+	 * as default
+	 * @return ConverterHub
+	 */
+	public static function getDefaultConverterHub($fresh = false) {
```
mrow4a, line 38: a bit hacky, but do not have better idea too

mrow4a (approved):

JRundfeldt:
@jvillafanez how long did you work on this?

jvillafanez:
The guid converter was already there, I just moved it. For the rest I think 2-3 days, including solution design, looking for the information, implementation and tests.

It doesn't cover all the possible binary attributes, but it should be easier to add support for them in case we need them.

JRundfeldt:
> The guid converter was already there, I just moved it. For the rest I think 2-3 days, including solution design, looking for the information, implementation and tests.
> 
> It doesn't cover all the possible binary attributes, but it should be easier to add support for them in case we need them.

Awesome, thx a lot! Can we have an estimate for release?

PR #770, opened 2022-12-14
Only trim on the string representation of an attribute

IljaN:
Trimming on binary could potentially remove control-characters which are actually part of the binary attribute value.


jvillafanez:
This should be good enough.
I don't think that the string representation of the binary value will contain characters that could be trimmed. We might want to be more strict though.

IljaN:
- [x] Changelog

IljaN:
I can't think up any scenario where a trim could "damage" the string representation.

pako81 (approved):

jvillafanez (approved):

Issue #797, opened 2023-06-27
Cannot detect base DN of an AD server when using an IP address


jnweiger:
Seen with user-ldap-0.17.0

This is a strange one.
I have an AD server running at a private network 10.42.0.2
with ldapsearch, I can happily connect like this:
```
ldapsearch -d 0 -H ldap://10.42.0.2 -D administrator@ker.jw-qa.owncloud.works -w "$password" -b dc=ker,dc=jw-qa,dc=owncloud,dc=works
```

In the web UI of user_ldap this IP address seems to work (Green bullet, connection OK)
But then the 'Detect Base DN' fails:
![image](https://github.com/owncloud/user_ldap/assets/1108546/11921032-b714-4a82-914d-e378b61cd928)

Server log snippet:
```
{"reqId":"CNKX0LO13BWW9dOexuSx","level":2,"time":"2023-06-27T10:06:11+00:00","remoteAddr":"","user":"--","app":"OCA\\User_LDAP\\User\\Manager","method":"--","url":"--","message":"No cached ldap result found for C4773B0D-A2C1-4970-9706-F9B28CD1F4DB"}
{"reqId":"CNKX0LO13BWW9dOexuSx","level":2,"time":"2023-06-27T10:06:11+00:00","remoteAddr":"","user":"--","app":"OCA\\User_LDAP\\User\\Manager","method":"--","url":"--","message":"No cached ldap result found for C4773B0D-A2C1-4970-9706-F9B28CD1F4DB"}
{"reqId":"CNKX0LO13BWW9dOexuSx","level":2,"time":"2023-06-27T10:06:11+00:00","remoteAddr":"","user":"--","app":"OCA\\User_LDAP\\User\\Manager","method":"--","url":"--","message":"No cached ldap result found for C4773B0D-A2C1-4970-9706-F9B28CD1F4DB"}
{"reqId":"TOD1j08AUXEU1HX5ul2x","level":1,"time":"2023-06-27T10:10:01+00:00","remoteAddr":"","user":"--","app":"cron","method":"--","url":"--","message":"Invalidating tokens older than 2023-06-27T09:50:01+00:00"}
[...]

PR #800, opened 2023-07-18
guessBaseDN: try to parse domain part from a user given in email syntax

jnweiger:
Suggested fix for #797

### DIFF
--- a/lib/Wizard.php
+++ b/lib/Wizard.php
@@ -685,6 +685,17 @@ class Wizard extends LDAPUtility {
 			}
 		}
 
+		//check whether the agent name is in email form
+		$i = \stripos($this->configuration->ldapAgentName, '@');
+		if ($i !== false) {
+			$base1 = \substr($this->configuration->ldapAgentName, $i+1);
+			$base1 = 'dc=' . \implode(',dc=', \explode('.', $base1));
+			if ($this->testBaseDN($base)) {
+				$this->applyFind('ldap_base', $base);
+				return $this->result;
+			}
+		}
+
 		//this did not help :(
 		//Let's see whether we can parse the Host URL and convert the domain to
 		//a base DN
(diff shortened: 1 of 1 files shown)


lib/Wizard.php
```diff
@@ -685,6 +685,17 @@ public function guessBaseDN() {
 			}
 		}
 
+		//check whether the agent name is in email form
+		$i = \stripos($this->configuration->ldapAgentName, '@');
+		if ($i !== false) {
+			$base1 = \substr($this->configuration->ldapAgentName, $i+1);
+			$base1 = 'dc=' . \implode(',dc=', \explode('.', $base1));
+			if ($this->testBaseDN($base)) {
+				$this->applyFind('ldap_base', $base);
```
jvillafanez, line 694: I guess this should be `$base1` instead of `$base`

lib/Wizard.php
```diff
@@ -685,6 +685,17 @@ public function guessBaseDN() {
 			}
 		}
 
+		//check whether the agent name is in email form
+		$i = \stripos($this->configuration->ldapAgentName, '@');
```
jvillafanez, line 689: `strrpos` should work better. We probably should get the position of the last "@" char in case there are multiple

jnweiger: good idea. checking the last never hurts!

lib/Wizard.php
jnweiger: oops, you are right, of course.
Thanks. pushed a fix.

lib/Wizard.php
```diff
@@ -685,6 +685,17 @@ public function guessBaseDN() {
 			}
 		}
 
+		//check whether the agent name is in email form; last '@' is taken, if muptiple.
```
jvillafanez, line 688: tiny typo, not `muptiple`

jvillafanez (approved):
I'm assuming it's tested and working

PR #801, opened 2023-08-09
Exposed attributes

jvillafanez:
Expose user attributes from LDAP to other apps.

The admin will expose some attributes for other apps to consume. For example, the admin could expose the `businessCategory`, `employeeType` and `employeeNumber` (assuming those attributes are available) so other apps can use that information.

By default, no attribute will be exposed. This means that other apps will have to use the information provided by core (check the core's `IUser` interface). The admin can add any attribute (one per line) in the user_ldap's wizard -> advanced tab -> special attributes -> exposed user attributes (at the bottom of the advanced tab).
Note that the information will be available for any app, so choose wisely what information you want to expose.

Other apps can access to those attributes via the `getExtendedAttributes` method of the `IUser` class (available since 10.11).
The recommended procedure to get the exposed LDAP attributes is:
1. Check that the current `IUser` instance is from LDAP (`$user->getBackendClassName() === 'LDAP'` should be enough)
2. Get the user's extended attributes: `$eattrs = $user->getExtendedAttributes()`
3. Check that the state of the LDAP attributes is OK. This is needed because we'll have to connect to the LDAP server and the connection could be down. `if (isset($eattrs['user_ldap_state']) && $eattrs['user_ldap_state'] === 'OK') ...`
    The `user_ldap_state` can be either `OK` if we get the attributes, or `Error` if something wrong happens. If it's `Error`, there won't be any LDAP attribute available.
    In addition, note that the `user_ldap_state` will appear only for LDAP users.
4. Once we know the state is OK, we can check whether the attribute we want is present.
    The attribute will have the following format: `user_ldap_attr_<lowercaseAttr>`, for example `user_ldap_attr_employeetype` or `user_ldap_attr_homephonenumber`

```
$user = $this->userManager->get('myUser');
if ($user->getBackendClassName() !== 'LDAP') {
  return; // abort, not LDAP user
}

$extendedAttrs = $user->getExtendedAttributes();
if (!isset($extendedAttrs['user_ldap_state']) || $extendedAttrs['user_ldap_state'] !== 'OK') {
  return; // abort because we couldn't get the LDAP attributes reliably
}

if (isset($extendedAttrs['user_ldap_attr_homephone'])) {
  printf("myUser's phone number is %s", $extendedAttrs['user_ldap_attr_homephone']);
} else {
  printf("myUser doesn't have a phone number");
}
```

Note that the information exposed isn't being used for now. It's expected that other apps document additional requirements if they need the admin to modify the user_ldap's configuration in order to expose specific attributes

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -102,7 +102,7 @@ class Application extends \OCP\AppFramework\App {
 
 		$uProxy = $container->query(User_Proxy::class);
 		$eventDispatcher = $server->getEventDispatcher();
-		$eventDispatcher->addListener(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, function(UserExtendedAttributesEvent $event) use ($uProxy) {
+		$eventDispatcher->addListener(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, function (UserExtendedAttributesEvent $event) use ($uProxy) {
 			$targetUser = $event->getUser();
 			if ($targetUser->getBackendClassName() !== 'LDAP') {
 				// If the user doesn't come from LDAP, there is nothing to do here.
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -25,6 +25,8 @@ use OCA\User_LDAP\Group_Proxy;
 use OCA\User_LDAP\Helper;
 use OCA\User_LDAP\LDAP;
 use OCA\User_LDAP\User_Proxy;
+use OCP\User\UserExtendedAttributesEvent;
+use OCA\User_LDAP\User\Manager;
 
 class Application extends \OCP\AppFramework\App {
 	/**
@@ -92,5 +94,31 @@ class Application extends \OCP\AppFramework\App {
 			$this->getContainer()->query(Helper::class),
 			'loginName2UserName'
 		);
+	}
+
+	public function registerEventListener() {
+		$container = $this->getContainer();
+		$server = $container->getServer();
+
+		$uProxy = $container->query(User_Proxy::class);
+		$eventDispatcher = $server->getEventDispatcher();
+		$eventDispatcher->addListener(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, function(UserExtendedAttributesEvent $event) use ($uProxy) {
+			$targetUser = $event->getUser();
+			if ($targetUser->getBackendClassName() !== 'LDAP') {
+				// If the user doesn't come from LDAP, there is nothing to do here.
+				return;
+			}
+
+			try {
+				$attrs = $uProxy->getExposedAttributes($targetUser->getUID());
+
+				$event->setAttributes('user_ldap_state', 'OK');
+				foreach ($attrs as $key => $value) {
+					$event->setAttributes("user_ldap_attr_{$key}", $value);
+				}
+			} catch (\Exception $e) {
+				$event->setAttributes('user_ldap_state', 'Error');
+			}
+		});
 	}
 }
--- a/lib/User/Manager.php
+++ b/lib/User/Manager.php
@@ -136,6 +136,22 @@ class Manager {
 	public function getConnection() {
 		return $this->access->getConnection();
 	}
+
+	/**
+	 * Get a list with the configured exposed attributes.
+	 * The `getAttributes` method will contain these exposed attributes, and all
+	 * of them will be requested. This list is public in order to know what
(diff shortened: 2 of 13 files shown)


jvillafanez:
Ugly patch for core in order to help testing this (don't use it in production)
```
diff --git a/settings/Panels/Personal/Profile.php b/settings/Panels/Personal/Profile.php
index bd7caeb062..3b626771da 100644
--- a/settings/Panels/Personal/Profile.php
+++ b/settings/Panels/Personal/Profile.php
@@ -108,6 +108,7 @@ class Profile implements ISettings {
                $tmpl->assign('languageSelector', $selector->fetchPage());
                $tmpl->assign('legal_privacy_policy', $this->config->getAppValue('core', 'legal.privacy_policy_url'));
                $tmpl->assign('legal_imprint', $this->config->getAppValue('core', 'legal.imprint_url'));
+               $tmpl->assign('user_extended_attributes', $this->userSession->getUser()->getExtendedAttributes());
                return $tmpl;
        }
 
diff --git a/settings/templates/panels/personal/profile.php b/settings/templates/panels/personal/profile.php
index 52ae176183..8aecfafaf0 100644
--- a/settings/templates/panels/personal/profile.php
+++ b/settings/templates/panels/personal/profile.php
@@ -138,6 +138,13 @@ if ($_['passwordChangeSupported']) {
 <?php
 }
 ?>
+
+       <ul>
+       <?php foreach ($_['user_extended_attributes'] as $attr => $value): ?>
+       <li><?php p($attr); ?> : <?php p($value); ?></li>
+       <?php endforeach; ?>
+       </ul>
+
 <form id="language" class="section">
        <h2>
                <label><?php p($l->t('Language'));?></label>
```
Screenshot from 2023-08-09 11-39-46

lib/User/Manager.php
```diff
@@ -136,6 +136,22 @@ private function checkAccess() {
 	public function getConnection() {
 		return $this->access->getConnection();
 	}
+
+	/**
+	 * Get a list with the configured exposed attributes.
+	 * The `getAttributes` method will contain these exposed attributes, and all
```
phil-davis, line 142: Not sure exactly what the comment means. Maybe
```suggestion
	 * The `getExposedAttributes` method will return an array of these exposed attributes, and all
```

lib/AppInfo/Application.php
```diff
@@ -93,4 +95,30 @@ public function registerHooks() {
...
+				// If the user doesn't come from LDAP, there is nothing to do here.
+				return;
+			}
+
+			try {
+				$attrs = $uProxy->getExposedAttributes($targetUser->getUID());
+
+				$event->setAttributes('user_ldap_state', 'OK');
+				foreach ($attrs as $key => $value) {
+					$event->setAttributes("user_ldap_attr_{$key}", $value);
```
phil-davis, line 117: This method is called `setAttributes` but actually it only sets one attribute.
But I see that `UserExtendedAttributesEvent` comes from core, so we can't adjust the method name now!

phil-davis (commented):
Looks OK. If the admin puts "rubbish" text into the textarea and saves, I wonder how a "reasonableness" check can be done to prevent saving obviously invalid crud?

lib/User/Manager.php
jvillafanez: The `getAttributes` method refers to the method below this one.
Basically, the `$manager->getExposedAttributes()` will return the attributes exposed by the admin, and the `getAttributes()` will return everything, including the exposed attributes. In other words, the `getAttributes` will contain the `getExposedAttributes`.

I guess I need to figure out how to explain it better.

lib/AppInfo/Application.php
jvillafanez: Yeah, I don't think we can do much to fix this. We might include another method in the event, but we'll have to check the ownCloud version here.

jvillafanez:
> Looks OK. If the admin puts "rubbish" text into the textarea and saves, I wonder how a "reasonableness" check can be done to prevent saving obviously invalid crud?

I don't think we can do anything. I don't know if we could access to the ldap schema, but it sounds painful. Without the schema, we can't know what attributes are available.
We could gather a list of all possible attributes, but we can't guarantee that there is no additional attribute outside of our list. Moreover, AD and openldap (for example) have a different set of attributes.
In the end, providing something that works reasonable well seems too problematic.

jvillafanez:
Minimum OC version for the app is already 10.11, which is the one we need for this feature to work, otherwise the event we need to listen to is missing and it could crash the app.

pako81:
Looks ok to me.

pako81 (approved):

PR #807, opened 2023-10-05
feat: php 8.3

DeepDiver1975:


DeepDiver1975:
needs https://github.com/owncloud-ci/core/pull/83

jvillafanez:
It seems some tests are failing because of a wrong replacement with the ldap port (there might be more, but the port seems to be the most common). The tests are getting the port as integer, which fails to be replaced because a string is expected. This is a problem with the tests, and it's unrelated to the actual code.

@phil-davis could you check it? the replacement seems to happen in core. We'll likely need changes there to fix this issue.

jvillafanez:
The rest of the failing tests at the moment are due to the federated server being OC10 (with php 7.4) trying to run on php 8.3, so it fails to install.
There are quite a bunch of tests, so skipping them doesn't look like a good idea. I'm not finding an easy switch to change the php version for the federated server... maybe QA has better ideas on how we can handle this in a good way.

phil-davis:
I think that we need https://github.com/owncloud/core/pull/41500

To fix this acceptance problem:
https://drone.owncloud.com/owncloud/user_ldap/5467/20/13
```
    When the administrator sets the LDAP config "LDAPTestId" key "ldapPort" to "%ldap_port%" using the occ command # UserLdapGeneralContext::ldapConfigHasKeySetTo()
      Type error: str_replace(): Argument #2 ($replace) must be of type array|string, int given (Behat\Testwork\Call\Exception\FatalThrowableError)
```

phil-davis:
https://github.com/owncloud/core/pull/41501 has been merged. That should fix a small test fail.

I will look more now at the federated setups, to get the correct PHP versions etc. for each server in each pipeline.

phil-davis (approved):
LGTM if CI passes

jvillafanez (approved):

PR #808, opened 2023-10-09
Changes for implementing the new sync mech from core

jvillafanez:
Implement new user sync mechanism developed in core (https://github.com/owncloud/core/pull/41016)

New feature is added conditionally, based on the presence of the new syncManager in core. This will keep compatibility with previous versions, so there is no need to rise the OC minimum supported version.


jvillafanez:
I guess the phpstan error will be fixed when the PR in core is merged, but I'm not sure if we should rise the minimum version. I don't think there will be problems if this code is run in 10.13 or earlier because the offending code won't be loaded, but maybe we should be more explicit and write a warning in the `UserSyncLDAPBackend` class.

Issue #809, opened 2023-10-10
[Acceptance] Nightly failure in `apiUserLDAP/sharingLocalUserLdapUser.feature:136`


grgprarup:
The acceptance test is failing in scenario `apiUserLDAP/sharingLocalUserLdapUser.feature:136` with `daily master` server
```feature
@issue-364
  Scenario: Share a folder from an LDAP user to a local user                                                                      # /var/www/owncloud/testrunner/apps/user_ldap/tests/acceptance/features/apiUserLDAP/sharingLocalUserLdapUser.feature:136
    Given user "Alice" has shared folder "/PARENT" with user "local-user"                                                         # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    When the administrator sets the LDAP config "LDAPTestId" key "ldapHost" to "not-existing" using the occ command               # UserLdapGeneralContext::ldapConfigHasKeySetTo()
    And user "local-user" downloads file "/PARENT (2)/parent.txt" using the WebDAV API                                            # FeatureContext::userDownloadsFileUsingTheAPI()
    Then the HTTP status code should be "500"                                                                                     # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 404 is not the expected value 500
      Failed asserting that 404 matches expected '500'.
    And the content of file "/PARENT (2)/parent.txt" for user "local-user" should be:                                             # FeatureContext::contentOfFileForUserShouldBePyString()
      """
      <?xml version="1.0" encoding="utf-8"?>
      <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
        <s:exception>InvalidArgumentException</s:exception>
        <s:message>Jail rootPath is null</s:message>
      </d:error>
      
      """
    When user "local-user" gets the properties of file "/PARENT (2)/parent.txt" using the WebDAV API                              # WebDavPropertiesContext::userGetsThePropertiesOfFolder()
    Then the HTTP status code should be "500"                                                                                     # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    When the administrator sets the LDAP config "LDAPTestId" key "ldapHost" to "%ldap_host_without_scheme%" using the occ command # UserLdapGeneralContext::ldapConfigHasKeySetTo()
    And user "local-user" gets the properties of file "/PARENT (2)/parent.txt" using the WebDAV API                               # WebDavPropertiesContext::userGetsThePropertiesOfFolder()
[...]

Issue #364, opened 2019-01-29
local user tries to access item shared by a LDAP user when LDAP server is offline


individual-it:
1. ldap user shares with local use
2. ldap server goes offline

local user cannot access the files
response:
```
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
<s:exception>InvalidArgumentException</s:exception>
<s:message>Jail rootPath is null</s:message>
</d:error>

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/user_ldap/issues/205 (Transfer users local to users ldap), https://github.com/owncloud/user_ldap/issues/51 (LDAP User A shares with LDAP User B. LDAP Server for user B leaves to be available. The permissions of the checkboxes should be disabled), https://github.com/owncloud/user_ldap/issues/294 (Can't remove LDAP users after removing the LDAP server), https://github.com/owncloud/user_ldap/issues/102 (Unable to see ldap users), and https://github.com/owncloud/user_ldap/issues/176 (ownCloud shares are removed temporarily when LDAP server is not available).

PrajwolAmatya:
With the latest fixes, the response now returns status code `404`. Is that the expected behavior?
cc @DeepDiver1975

phil-davis:
core PR  https://github.com/owncloud/core/pull/41059 reverted the change in core.
Now this issue is happening again like it used to.
So we have a failing test in `user_ldap` CI. See issue https://github.com/owncloud/user_ldap/issues/814

PR #811, opened 2023-10-10
[tests-only][full-ci] update status code from 500 to 404

PrajwolAmatya:
In this PR, the status code is updated from `500` to `404`, for `local-user` trying to download the file content shared by LDAP user.

Fixes: https://github.com/owncloud/user_ldap/issues/809


phil-davis (approved):
That will be great if CI passes.
The server has been fixed somewhere so that a proper status code is returned rather than a 500.

phil-davis:
It seems that core master (and 10.13.2) has fixed user_ldap issue #364 
We can double-check that, and close if it is fixed.

I skipped the adjusted test scenario on older oC10 core versions, because the fix/change-in-behavior seems to have happened with core 10.13.2 and core master.

Please merge when CI passes.

PR #813, opened 2023-10-11
test: add IntegrationTestPagingMultipleBase to drone

DeepDiver1975:


phil-davis:
@DeepDiver1975 I don't think that the drone code is smart enough to realize that changes to `.drone.star` mean that all tests (potentially) need to be run. If there is no change to real code, or actual test code, then it skips lots of test steps.

Put `full-ci` in the title, and the drone code "knows" to run everything, regardless of the files that are touched in the PR.

I edited the PR title. To get drone to "see" the change in the title, you have to push some different commit to the branch (add a new commit, or rebase and force push so that there is a different commit id)

DeepDiver1975:
test was executed and is green -> merge

image

Issue #814, opened 2023-11-06
Nightly failure: apiUserLDAP/sharingLocalUserLdapUser.feature:136


saw-jan:
Build: https://drone.owncloud.com/owncloud/user_ldap/4804/29/12, https://drone.owncloud.com/owncloud/user_ldap/4804/37/12

```feature
  @issue-364 @skipOnOcV10.12 @skipOnOcV10.13.0 @skipOnOcV10.13.1
  Scenario: Share a folder from an LDAP user to a local user                                                                      # /var/www/owncloud/testrunner/apps/user_ldap/tests/acceptance/features/apiUserLDAP/sharingLocalUserLdapUser.feature:136
    Given user "Alice" has shared folder "/PARENT" with user "local-user"                                                         # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    When the administrator sets the LDAP config "LDAPTestId" key "ldapHost" to "not-existing" using the occ command               # UserLdapGeneralContext::ldapConfigHasKeySetTo()
    And user "local-user" downloads file "/PARENT (2)/parent.txt" using the WebDAV API                                            # FeatureContext::userDownloadsFileUsingTheAPI()
    Then the HTTP status code should be "404"                                                                                     # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 500 is not the expected value 404
      Failed asserting that 500 matches expected '404'.
```

saw-jan:
Similar failure https://github.com/owncloud/files_lifecycle/issues/555

phil-davis:
core PR revert has caused this. I have asked about it in https://github.com/owncloud/core/pull/41059#issuecomment-1794164444

phil-davis:
Let's revert the tests so that we get green CI that demonstrates the behavior that currently exists.

PR #815, opened 2023-11-21
fix: search terms for binary fields will be converted if possible

jvillafanez:
Related to https://github.com/owncloud/enterprise/issues/6032

The app will take care of converting binary attributes when searching for them.

For example, if the user_ldap app has been configured the "user search attributes" with "objectguid" (among other attributes), searching by "36E271F3" might return results, assuming the converted value we have from binary contains those characters.
Note that the converted value will also be stored as search term for the account in the "oc_account_terms" table.

### DIFF
--- a/lib/Access.php
+++ b/lib/Access.php
@@ -42,6 +42,7 @@ use OCA\User_LDAP\User\IUserTools;
 use OCA\User_LDAP\User\Manager;
 use OCA\User_LDAP\Mapping\AbstractMapping;
 use OCA\User_LDAP\Attributes\ConverterHub;
+use OCA\User_LDAP\Attributes\ConverterException;
 use OCP\Util;
 
 /**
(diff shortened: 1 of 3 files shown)


jvillafanez:
Failing tests are known: https://github.com/owncloud/user_ldap/issues/814

jvillafanez:
I don't think I can add tests for the changes in the `Access` class easily. Need to change attributes in the connection class, which are accessed using magic methods. It doesn't seem possible to mock those.

pako81:
merge this?

dj4oC (commented):
Adds binary-attribute conversion (via `ConverterHub`) into the LDAP search-filter builders and `UserEntry::getSearchTerms()`, so binary attributes like `objectguid` can be searched by their string form.

### Findings

**Security/correctness — needs human verification before merge (LDAP attribute mapping is a never-auto-fix zone):**
- In both `getAdvancedFilterPartForSearch()` and `getFilterPartForSearch()`, the converted value is escaped a second time: `$this->prepareSearchTerm($converterHub->str2filter($attr, $word))`. If `str2filter()` already returns an LDAP-filter-safe encoded form (e.g. a `\hh\hh\hh`-style hex escape for binary GUIDs, which is the standard way to filter on binary attributes per RFC 4515), then running that through a generic `prepareSearchTerm()` escaper again risks double-escaping the backslashes and corrupting the filter — likely breaking the search silently (filter would just never match) rather than being exploitable, but it should be confirmed against a real LDAP directory/mock, not just the unit test in this diff, which only exercises `bin2str()` on the read path (`UserEntryTest::testGetSearchTermsWithConversion`), not the filter-building path with the double-escape.
- No test covers `getAdvancedFilterPartForSearch()`/`getFilterPartForSearch()` with a converter-backed attribute — the one new test only covers `getSearchTerms()`.

**Stability:** the `catch (ConverterException $e)` fallback paths look correct (fall back to the unconverted, still-escaped term rather than failing the whole search).

### Verdict
Commenting, not blocking outright — but per policy, LDAP attribute-conversion/mapping changes need a maintainer to confirm the filter round-trips correctly against a real directory (or a filter-level test) before merge, given the double-escape risk above.

---

PR #823, opened 2024-08-30
fix: do no hit LDAP if no exposed attribute is configured

jvillafanez:
If there are no exposed attributes configured in the ldap's wizard, we shouldn't hit the LDAP server trying to get the exposed attributes.

Link: https://github.com/owncloud/enterprise/issues/6856


IljaN (approved):

jvillafanez:
good to merge now

Issue #824, opened 2024-09-27
Nightly build failure in `phpstan` pipeline


Salipa-Gurung:
Nightly build failed in `phpstan` of `phpstan-php7.4` pipeline.
```
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=4G --configuration=./phpstan.neon --no-progress --level=5 appinfo lib
 ------ ---------------------------------------------------------------------- 
  Line   lib/Wizard.php                                                        
 ------ ---------------------------------------------------------------------- 
         Ignored error pattern #Strict comparison using === between mixed and  
         '' will always evaluate to false.# in path                            
         /var/www/owncloud/server/apps/user_ldap/lib/Wizard.php was not        
         matched in reported errors.                                           
  866    Strict comparison using === between                                   
         mixed~(0|0.0|''|'0'|array{}|false|null) and '' will always evaluate   
         to false.                                                             
         🪪  identical.alwaysFalse                                              
 ------ ---------------------------------------------------------------------- 


 [ERROR] Found 2 errors
```
Build: https://drone.owncloud.com/owncloud/user_ldap/5207/2/4

PR #825, opened 2024-09-27
fix: strict comparison

saw-jan:
Fixes https://github.com/owncloud/user_ldap/issues/824


phil-davis (approved):
LGTM - the code explicitly checks for the empty string, as well as false, so this should be fine.

Issue #49, opened 2017-02-10
Cannot log into ownCloud Server when one of the LDAP Servers written in the wizard have a wrong credentials


davitol:
### Steps to reproduce
1. As admin enable LDAP app
2. In LDAP section enable two LDAP configurations
3. Make one of the LDAP server backend not accesible changing the data written in the LDAP wizard with a wrong one
4. Try to log in with a user from the other LDAP Server (aaliyah_beer in this case)

### Expected behaviour

The users that belongs to the other LDAP server should be able to log in

### Actual behaviour

No user can log into the oC server. Error is shown

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
5.6

**ownCloud version:**

"9.1.3.1","versionstring":"9.1.3","edition":"Enterprise"

**Updated from an older ownCloud or fresh install:**
Fresh

**Are you using external storage, if yes which one:** local/smb/sftp/...
No

**Are you using encryption:** 
No

**Logs**

```
[...]

PR #828, opened 2025-01-24
fix: return null instead of false in case of invalid quota

jvillafanez:
Prevent switching to a different server (in a multiple server setup) in case quota isn't set for a particular user.

Related to https://github.com/owncloud/enterprise/issues/7040

### DIFF
--- a/lib/User/UserEntry.php
+++ b/lib/User/UserEntry.php
@@ -210,7 +210,7 @@ class UserEntry {
 	 * values are 'none' (unlimited), 'default' (the ownCloud's default quota), '1234' (quota in
 	 * bytes), '1234 MB' (quota in MB - check the \OC_Helper::computerFileSize method for more info)
 	 *
-	 * @return string|false quota
+	 * @return string|null quota
 	 * TODO throw Exception for invalid values after https://github.com/owncloud/core/pull/28805 has been merged
 	 */
 	public function getQuota() {
@@ -237,9 +237,6 @@ class UserEntry {
 					$quota = null;
 				}
 			}
-		}
-		if ($quota === null) {
-			$quota = false;
 		}
 
 		\OC::$server->getLogger()->debug("using quota <$quota> for user <{$this->getOwnCloudUID()}>", ['app' => 'user_ldap']);


phil-davis:
There are some test fails for tests that seem to be better now.
For example, a test scenario linked to issue #49 
https://drone.owncloud.com/owncloud/user_ldap/5351/29/13
```
  Scenario: authentication works when second of multiple configurations has an unreachable host configured # /var/www/owncloud/testrunner/apps/user_ldap/tests/acceptance/features/apiUserLDAP/serverConnection.feature:54

    Given a new LDAP config with the name "wrongLdapConfig" has been created                               # UserLdapGeneralContext::createNewLdapConfig()
    And LDAP config "wrongLdapConfig" has these settings:                                                  # UserLdapGeneralContext::ldapConfigHasTheseSettings()
...

    When user "Alice" requests "/index.php/apps/files" with "GET" using basic auth                         # AuthContext::userRequestsURLWithUsingBasicAuth()
    Then the HTTP status code should be "500"                                                              # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 200 is not the expected value 500
      Failed asserting that 200 matches expected '500'.
```

The test can be adjusted to expect the now good 200 status.

jvillafanez:
But the "authentication works when first of multiple configurations has an unreachable host configured" responds with a 500 status. This seems inconsistent to me, and people will complain.
Right now, you need to have all the LDAP servers connected in order to log in. It might not be the best idea, but it works like that.

dj4oC (commented):
Changes `UserEntry::getQuota()` to return `null` instead of `false` for an invalid/unset quota value.

### Findings

**Stability — potential backward-compat break:** this changes the documented return contract from `string|false` to `string|null`. Any caller (in this app or in `owncloud/core`) doing a strict `=== false` (or `== false`, which would still work) check on the result of `getQuota()`/wherever this value flows into `IUser::setQuota()` or similar would silently stop treating "no quota set" as the sentinel it expects. I don't have visibility into all call sites from this diff alone — please confirm no caller relies on `=== false` specifically before merging (a plain `grep -rn "getQuota()" ` across user_ldap + core call sites would settle it).
- The change itself (aligning with PHP's more idiomatic `null` for "no value") is reasonable and the removed `if ($quota === null) { $quota = false; }` block is consistent with the new signature.

**Test coverage:** no test included for the new null-return path.

### Verdict
Commenting — low risk, but non-blocking confirmation needed on caller compatibility and a regression test for the null path before merge.

---

PR #844, opened 2026-07-22
fix: strip CRLF/control bytes from LDAP bind DN

DeepDiver1975:
## Summary
- The LDAP bind DN (`ldap_dn` / `ldapAgentName`, settable via the wizard ajax endpoint) was only `trim()`-ed, unlike `ldapAgentPassword` which already strips control characters for the same reason (see 80f213ff).
- Since the bind DN is sent over the same `ldap_bind()` call to a host/port the admin fully controls, an unsanitized CRLF allowed raw protocol bytes to be smuggled to whatever service is actually listening there — reported as an SSRF chained into RCE (OC10-134).
- Fix: route `ldapAgentName` through the same `FILTER_UNSAFE_RAW` / `FILTER_FLAG_STRIP_LOW` filter already applied to `ldapAgentPassword`.

## Test plan
- [x] Added a regression test asserting CRLF/control bytes are stripped when setting `ldapAgentName` (`tests/unit/ConfigurationTest.php`)
- [x] Confirmed the new test fails against the old code and passes with the fix
- [x] Ran the full `user_ldap` unit suite (224 tests) — all green, no regressions

Ref: OC10-134

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased] - XXXX-XX-XX
 
+### Fixed
+
+- [#844](https://github.com/owncloud/user_ldap/pull/844) - Strip control characters from the LDAP bind DN and the search filters
+
 
 ## [0.20.1] - 2026-07-22
 
--- a/lib/Configuration.php
+++ b/lib/Configuration.php
@@ -244,6 +244,7 @@ class Configuration {
 
 			$setMethod = 'setValue';
 			switch ($key) {
+				case 'ldapAgentName':
 				case 'ldapAgentPassword':
 					$val = \filter_var($val, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW);
 					$setMethod = 'setRawValue';
--- a/lib/Configuration.php
+++ b/lib/Configuration.php
@@ -245,7 +245,20 @@ class Configuration {
 			$setMethod = 'setValue';
 			switch ($key) {
 				case 'ldapAgentName':
+				case 'ldapUserFilter':
+				case 'ldapLoginFilter':
+				case 'ldapGroupFilter':
+					// These end up verbatim in the bind request resp. in the
+					// search filter sent over the wire. Raw control characters
+					// have no legitimate meaning here (RFC 4515 requires them
+					// to be escaped) and would allow smuggling protocol bytes
+					// to whatever service listens on the configured host:port.
+					// Strip them, but keep the regular trimming of setValue().
+					$val = \filter_var($val, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW);
+					break;
 				case 'ldapAgentPassword':
+					// same reasoning as above, but the password must not be
+					// trimmed since leading/trailing whitespace is significant
 					$val = \filter_var($val, FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW);
 					$setMethod = 'setRawValue';
 					break;
--- a/tests/unit/ConfigurationTest.php
+++ b/tests/unit/ConfigurationTest.php
@@ -68,6 +68,9 @@ class ConfigurationTest extends \Test\TestCase {
 
 		$password = ' such a passw0rd ';
 
+		$dnWithCrlf = "cn=admin\r\nset foo bar\r\n,dc=example,dc=org";
+		$expectedDn = 'cn=adminset foo bar,dc=example,dc=org';
+
 		return [
 			'set general base' => ['ldapBase', $inputWithDN, $expectWithDN],
 			'set user base'    => ['ldapBaseUsers', $inputWithDN, $expectWithDN],
@@ -83,6 +86,7 @@ class ConfigurationTest extends \Test\TestCase {
 			'set login filter attributes'    => ['ldapLoginFilterAttributes', $inputNames, $expectedNames],
 
 			'set agent password' => ['ldapAgentPassword', $password, $password],
+			'set agent name strips CRLF' => ['ldapAgentName', $dnWithCrlf, $expectedDn],
 
 			'set home folder, variant 1' => ['homeFolderNamingRule', $inputHomeFolder[0], $expectedHomeFolder[0]],
 			'set home folder, variant 2' => ['homeFolderNamingRule', $inputHomeFolder[1], $expectedHomeFolder[1]],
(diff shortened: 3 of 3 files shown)


DeepDiver1975:
Closing — this fix needs to go through the private security-advisory process rather than a public PR. Continuing under OC10-134 / a private advisory.

jvillafanez (dismissed):

DeepDiver1975:
## Code review

**Scope reviewed:** `lib/Configuration.php` + `tests/unit/ConfigurationTest.php` (+5/-0 as originally pushed).

The approach is right: the bind DN is admin-settable via the wizard AJAX `save` action and `occ ldap:set-config`, all write paths funnel through `setConfiguration()`, and reusing the 2021 password mitigation (80f213ff) is the consistent call. Nice side effect: `Access::areCredentialsValid()` stuffs the user's DN into `ldapAgentName`, so that path is covered too.

Two things needed fixing, both pushed in b8b9cfc6.

### 1. Unintended behaviour change: the bind DN was no longer trimmed

Sharing the `case` with the password also inherited `$setMethod = 'setRawValue'`, which skips the `trim()` that `setValue()` applies. `FILTER_FLAG_STRIP_LOW` strips `0x00–0x1F` but **not** `0x20`, so:

* A copy-pasted `" cn=admin,dc=example,dc=org "` kept its trailing space and got sent to `ldap_bind()` verbatim, where previously `trim()` silently repaired it.
* A whitespace-only DN no longer normalised to `''`, so `Connection::validateConfiguration()` tripped the "either no password is given for the user agent…" branch and `ConfigurationController::test()` reported the config invalid.

The raw setter exists specifically so *passwords* keep significant leading/trailing whitespace (cf. the 0.17.0 changelog note); a DN has no such requirement. The DN now has its own `case` that falls through to `setValue`, keeping the previous trim semantics while still stripping the dangerous bytes.

### 2. The mitigation was narrower than the bug class

If CRLF in the bind DN is exploitable over `ldap_bind()`, CRLF in a **search filter** is equally exploitable over `ldap_search()` on the same connection — and searches are reachable on ordinary login / user-list requests, not only on an explicit "test connection" click. `ldapUserFilter`, `ldapLoginFilter` and `ldapGroupFilter` went through plain `setValue` (trim only) and land in filter strings on the wire. Those three now get the same `FILTER_FLAG_STRIP_LOW` treatment. `ldapBase*` and the `*Objectclass`/`*Groups` lists were already safe, incidentally, because `setMultiLine()` splits on `\r\n`.

### Smaller notes

* **Title/description overstate the filter.** `STRIP_LOW` covers `< 0x20` only — `0x7F` (DEL) survives (`filter_var("cn=a\x7fb", …)` → `cn=a\x7fb`). Harmless for CRLF smuggling, but worth not claiming more than it delivers. `FILTER_FLAG_STRIP_HIGH` is correctly *not* used; it would mangle UTF-8 DNs, and there's now a test pinning that.
* **Array input silently becomes `false`.** `filter_var([...], FILTER_UNSAFE_RAW, …)` returns `false`, so `ldap_dn[]=x` now stores boolean `false` instead of passing through. Pre-existing quirk shared with the password key — left as is, just noting the sanitizer assumes a string.
* **Changelog was missing** for a security fix; added under `## [Unreleased]` → `### Fixed`.
* Added an inline comment explaining *why* the stripping exists — that knowledge previously lived only in a 2021 commit message, which is why it had to be rediscovered here.

### Tests

The original row was idiomatic and slots into the existing provider correctly. The gap that let #1 through was the absence of a case pinning the trim behaviour, so the follow-up adds: DN trimming, UTF-8 DN survival, and CRLF stripping for each of the three filters. All five new expectations fail against the pre-follow-up code and pass after.

Verification: I could not run the suite through `phpunit.xml` in my checkout (its bootstrap needs a live core DB), so I exercised the real `Configuration` class directly against a stub `IConfig` — 8/8 expectations pass, and 5/8 fail when the fix is stashed. `php-cs-fixer` reports 0 of 97 files needing changes. CI remains the authority on the full 224-test run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

DeepDiver1975:
## Security review — no vulnerabilities found

Focused security review of the two commits on this branch (`lib/Configuration.php`). **No findings at high confidence.** The change is a net security improvement.

### Write path is closed

All writes to `ldapAgentName`, `ldapUserFilter`, `ldapLoginFilter` and `ldapGroupFilter` funnel through `Configuration::setConfiguration()` — the app has exactly one `setAppValue()` call (`lib/Configuration.php:513`). Reachable writers are the admin-only wizard/config controllers (no `@NoAdminRequired` anywhere in `lib/Controller/`), `Wizard::applyFind()` (`lib/Wizard.php:730`), the `occ` commands, and `Access::areCredentialsValid()` (`lib/Access.php:1654`, which sets the DN on an in-memory cloned connection and never persists it). So the new sanitization cannot be bypassed, and no privilege boundary is crossed by the behaviour change.

### Hypotheses tested and ruled out

- **`filter_var()` returns `false` for arrays.** Only reachable if an admin POSTs `cfgval[]=…` for one of these four keys; they are never arrays on any other path (array-valued keys use `setMultiLine`). It fails *closed*: `ldapLoginFilter === false` makes `Connection::doCriticalValidation()` fail and `establishConnection()` refuse to connect. Pre-PR the same input stored a raw array and was equally broken. Admin-self-inflicted, no auth impact.
- **Trim regression from the first commit.** Real, and correctly fixed by b8b9cfc — the DN gets its own case and keeps `setValue()`'s `trim()`, so a whitespace-only DN still normalizes to `''` and the agent/password pairing check in `doCriticalValidation()` still fires. Pinned by the new trim test.
- **`null` DN via `areCredentialsValid()`.** `UserEntry::getDN()` can return `null`; post-PR that becomes `''`, which now trips the `$agent === '' && $pwd !== ''` guard (`lib/Connection.php:473`) instead of reaching `ldap_bind($link, null, $password)` — an RFC 4513 unauthenticated bind. Hardening, not a regression.
- **Stripping as a bypass of a raw-value validation.** The only value-level checks are `empty()` and `mb_strpos($filter, '%uid')`. Stripping can only *create* a `%uid` placeholder, never remove one, and both checks are usability guards on admin-authored input, not security boundaries.
- **Stripping joining fragments into a wider filter.** Cannot widen the trust model — the admin can already author any filter.
- **UTF-8 mangling.** `FILTER_FLAG_STRIP_HIGH` deliberately absent; only bytes `< 0x20` are stripped, so non-ASCII DNs survive intact and no bind can be silently retargeted. Pinned by the umlaut test.
- **`switch` fall-through.** The new case block `break`s before `case 'ldapAgentPassword'`, so the password keeps `setRawValue` and the DN/filters keep `setValue`.

### Informational — out of scope for this PR

This closes the admin-authored half of the class but not the user-controlled half: `lib/Access.php:735-736` and `:749-750` apply `escapeFilterPart($loginName)` (which escapes only `* \ ( )`, see `Access.php:1467-1476`) and then `str_replace('%uid', $loginName, …)`, so control bytes in a *login name* still reach `ldap_search()`. Pre-existing and untouched here, and no code newly relies on the config-level stripping as a guarantee — but worth knowing the other half stays open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jvillafanez (approved):

